PR #2, opened 2020-11-27
Add empty dict as default for updateStrategy (Fix #1)

### LINKED ISSUE #1: coalesce.go:160: warning: skipped value for updateStrategy: Not a table.
I tried to upgrade from stable repo to this chart, but I get a warning:

```sh
coalesce.go:160: warning: skipped value for updateStrategy: Not a table.
```

The chart gets upgraded, but I cannot verify whether the updateStrategy was honored. Did anything change in this chart compared to stable?

This is the updateStragety from my values.yaml:

```yaml
updateStrategy:                                                                 
  type: Recreate       
```

petarGitNik:
This should fix issue #1. Fix is as suggested in the comment [here][1]. Let me know if I need to update anything else, such as changelog or should I bump the version for the next release.

[1]: https://github.com/twuni/docker-registry.helm/issues/1#issuecomment-734377844

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -3,7 +3,7 @@
 # Declare variables to be passed into your templates.
 replicaCount: 1
 
-updateStrategy:
+updateStrategy: {}
   # type: RollingUpdate
   # rollingUpdate:
   #   maxSurge: 1


canterberry (approved):
:+1: LGTM. Minimal and direct.

> :bulb: Looks like I need to update the CI configuration to run on PRs as well.

PR #6, opened 2020-12-01
Syseleven/add prometheus support

Arabus:
Thought this might be useful ;-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,13 @@ their default values.
 | `ingress.tls`               | Ingress TLS configuration (YAML)                                                           | `[]`            |
 | `extraVolumeMounts`         | Additional volumeMounts to the registry container                                          | `[]`            |
 | `extraVolumes`              | Additional volumes to the pod                                                              | `[]`            |
+| `metrics.enabled`           | Enable the debug metrics interface of the registry                                         | `false`         |
+| `metrics.servicePortName    | The name of the service port for the debug interface                                       | `debug`         |
+| `metrics.servicePort        | The port number of the service port for the debug interface                                | `5001`          |
+| `metrics.targetPortName     | The name of the container port of the debug interface                                      | `debug`         |
+| `metrics.serviceMonitor     | Various Prometheus serviceMonitor resource configs                                         | `nil`           |
+| `metrics.serviceMonitor.enabled` | Deploy a prometheus serviceMonitor                                                    | `false`         |
+| `metrics.serviceMonitor.path` | The debug interface path for prometheus metrics. Adjust this if you set it via configData.http.debug.prometheus.path | `/metrics` |
 
 Specify each parameter using the `--set key=value[,key=value]` argument to
 `helm install`.
--- a/templates/deployment.yaml
+++ b/templates/deployment.yaml
@@ -13,10 +13,10 @@ spec:
       app: {{ template "docker-registry.name" . }}
       release: {{ .Release.Name }}
   replicas: {{ .Values.replicaCount }}
-{{- if .Values.updateStrategy }}
+  {{- if .Values.updateStrategy }}
   strategy:
-{{ toYaml .Values.updateStrategy | indent 4 }}
-{{- end }}
+    {{- toYaml .Values.updateStrategy | nindent 4 }}
+  {{- end }}
   minReadySeconds: 5
   template:
     metadata:
@@ -24,26 +24,26 @@ spec:
         app: {{ template "docker-registry.name" . }}
         release: {{ .Release.Name }}
         {{- if .Values.podLabels }}
-{{ toYaml .Values.podLabels | indent 8 }}
+          {{- toYaml .Values.podLabels | nindent 8 }}
         {{- end }}
       annotations:
         checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
-{{- if $.Values.podAnnotations }}
-{{ toYaml $.Values.podAnnotations | indent 8 }}
-{{- end }}
+        {{- if $.Values.podAnnotations }}
+          {{- toYaml $.Values.podAnnotations | nindent 8 }}
+        {{- end }}
     spec:
       {{- if .Values.imagePullSecrets }}
       imagePullSecrets:
-{{ toYaml .Values.imagePullSecrets | indent 8 }}
+        {{- toYaml .Values.imagePullSecrets | nindent 8 }}
       {{- end }}
-{{- if .Values.priorityClassName }}
+      {{- if .Values.priorityClassName }}
       priorityClassName: "{{ .Values.priorityClassName }}"
-{{- end }}
-{{- if .Values.securityContext.enabled }}
+      {{- end }}
+      {{- if .Values.securityContext.enabled }}
       securityContext:
         fsGroup: {{ .Values.securityContext.fsGroup }}
         runAsUser: {{ .Values.securityContext.runAsUser }}
-{{- end }}
+      {{- end }}
       containers:
         - name: {{ .Chart.Name }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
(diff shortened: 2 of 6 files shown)


templates/service.yaml
```diff
@@ -30,6 +30,12 @@ spec:
 {{- if (and (eq .Values.service.type "NodePort") (not (empty .Values.service.nodePort))) }}
       nodePort: {{ .Values.service.nodePort }}
 {{- end }}
+      {{- if .Values.serviceMonitor.enabled }}
+    - port: {{ (split ':' .Values.configData.http.debug.addr)._1 }}
```
canterberry, line 34: If we're discarding the IP address portion of `configData.http.debug.addr`, let's just rename `addr` to `port` and have that expect the port alone. Then, this line could just be `- port: {{ .Values.configData.http.debug.port }}`.

templates/servicemonitor.yaml
```diff
@@ -0,0 +1,20 @@
+{{ if .Values.serviceMonitor.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ include "docker-registry.fullname" . }}
+  labels:
+    {{- include "docker-registry.labels" . | nindent 4 }}
```
canterberry, line 7: :+1: I like this -- much cleaner than the hard-coded/duplicated labels in each of the other templates. Can this be extracted into a separate pull request, though? If this PR introduces a `docker-registry.labels` helper, I'd like to simultaneously adopt that helper consistently across all templates instead of applying to just one.

templates/servicemonitor.yaml
```diff
@@ -0,0 +1,20 @@
...
+spec:
+  endpoints:
+    - port: debug
+      path: /metrics
```
canterberry, line 11: Does it make sense for this path to be configurable? Or is the path here hard-coded based on what the container already provides? Are there ways to configure this chart that might make the path invalid?

values.yaml
```diff
@@ -107,6 +107,12 @@ configData:
     addr: :5000
     headers:
       X-Content-Type-Options: [nosniff]
+    # Enable this if you want prometheus metrics
+    #debug:
+    #  addr: 0.0.0.0:5001
+    #  prometheus:
+    #    enabled: true
+    #    path: /metrics
```
canterberry, line 115: `configData.http.debug.prometheus` is commented-out here but is not referenced anywhere in the templates.

templates/servicemonitor.yaml
```diff
@@ -0,0 +1,20 @@
+{{ if .Values.serviceMonitor.enabled }}
```
canterberry, line 1: The service monitor requires a metrics endpoint to be configured and available. This should not only check whether the chart is configured to enable the service monitor, but also whether there is anything to monitor. Basically, this template could use a `.Values.configData.http.debug.prometheus.enabled` check as well.

canterberry:
Thanks @Arabus! Initial review pass complete. Thanks for this.
As I got immersed more deeply in review, it all started to fall into place, and some comments I had toward the beginning of the review I was able to delete. What's left is distilled to a few spots of minor feedback and adjustments.

To reiterate my inline comment regarding the new helpers: I like where you're going with those changes, but I do think it muddies up this PR a bit and deserves its own PR to tackle separately and adopt consistently for other templates as well, not just the new servicemonitor.yaml template. For now, just follow the labels/conventions adopted in the other templates, dated as they may be. Then we can address modernization as a whole in a separate thread.

values.yaml
```diff
@@ -145,3 +151,8 @@ extraVolumes: []
 #        - key: cloudfront.pem
 #          path: cloudfront.pem
 #          mode: 511
+
+serviceMonitor:
+  enabled: false
```
canterberry, line 156: :memo: Need to document this new configuration value in the README.

templates/servicemonitor.yaml
Arabus: Sounds reasonable - this was mainly part of my PR because I already used it in a helm chart I had written for the same purpose as this ( which I wrote because I thought this one had been abandoned). I can't claim credit for that though, it is a standard feature of modern helm charts when building them via `helm create`. 

I will extract it for the time being and create a separate PR based on on this one.

templates/servicemonitor.yaml
Arabus: Integrated this, though checking for values in optional subdicts in helm is pita. Also setting global values from inside an if feels messy.

templates/servicemonitor.yaml
Arabus: This was rather tricky because the complete `http.` subdict is optional and helm tends to fail with nilPointer exceptions whe you reference values whose parents aren't there. I found a workaround though i am not really happy with it:
`{{- if not ( and ( hasKey .Values "configData" ) ( hasKey .Values.configData "http" ) ( hasKey .Values.configData.http "debug" ) ( hasKey .Values.configData.http.debug "prometheus" ) ( hasKey .Values.configData.http.debug.prometheus "enabled" ) ( .Values.configData.http.debug.prometheus.enabled ) ) -}}`

values.yaml
Arabus: Check, also added some lines to the values.yaml

templates/servicemonitor.yaml
Arabus: scratch that ... go and does not use lazy evaluation

Arabus: ok i blew a whole working day on this and it appears there is no way to make helm lookup a nested dict value where the parent dict does not exist... The only viable solution seems to be stacking if clauses... i will put them in tommorow after i have gone and yelled at the helm devs a little...

canterberry: I think `dig` might be what you need here. http://masterminds.github.io/sprig/dicts.html

Arabus: Holy cow... I scrolled through the libraries the whole day and missed that one completely... ok that should solve the above approach. I have gone and rewritten the PR in the meantime though, going with a more "user centric" one and allowing people to enable the debug port / metrics service endpoint without enabling the serviceMonitor at all. 

I am currently finishing it a little more and will push it then.

Arabus: ok updated the PR - this has become somewhat of a code splurge so if this feels like too much at once I am happy to break it down into smaller PRs.

README.md
```diff
@@ -86,6 +86,13 @@ their default values.
 | `ingress.tls`               | Ingress TLS configuration (YAML)                                                           | `[]`            |
 | `extraVolumeMounts`         | Additional volumeMounts to the registry container                                          | `[]`            |
 | `extraVolumes`              | Additional volumes to the pod                                                              | `[]`            |
+| `metrics.enabled`           | Enable the debug metrics interface of the registry                                         | `false`         |
+| `metrics.servicePortName    | The name of the service port for the debug interface                                       | `debug`         |
+| `metrics.servicePort        | The port number of the service port for the debug interface                                | `5001`          |
+| `metrics.targetPortName     | The name of the container port of the debug interface                                      | `debug`         |
+| `metrics.serviceMonitor     | Various Prometheus serviceMonitor resource configs                                         | `nil`           |
```
canterberry, line 93: Missing closing backticks on these four lines.

templates/deployment.yaml
```diff
@@ -54,46 +54,54 @@ spec:
           - /etc/docker/registry/config.yml
           ports:
             - containerPort: 5000
+            {{- if or .Values.metrics.enabled .Values.configData.http.debug.addr }}
```
canterberry, line 57: `or` should be `and` here.

templates/deployment.yaml
```diff
@@ -54,46 +54,54 @@ spec:
...
+            - name: {{ .Values.metrics.targetPortName | default "debug" | quote }}
+              containerPort: {{ (split ":" .Values.configData.http.debug.addr)._1 }}
+            {{- else }}
+              {{- if .Values.metrics.serviceMonitor.enabled }}
+                {{- fail "You enabled the serviceMonitor without enabling the registry debug endpoint. Please set metrics.enabled: true or specify configData.http.debug.addr" }}
```
canterberry, line 62: "or" should be "and" here as well.

Basically, `metrics.enabled` only makes sense if the registry is configured to emit metrics, so it effectively depends on `configData.http.debug.addr` being set.

templates/servicemonitor.yaml
```diff
@@ -0,0 +1,97 @@
...
+    matchNames:
+      - {{ .Release.Namespace | quote }}
+  {{- if .Values.metrics.serviceMonitor.sampleLimit }}
+  sampleLimit: {{ .Values.metrics.serviceMonitor.sampleLimit }}
+  {{- end }}
+  {{- if .Values.metrics.serviceMonitor.targetLimit }}
+  targetLimit: {{ .Values.metrics.serviceMonitor.targetLimit }}
+  {{- end }}
+{{- end }}
+{{- end }}
```
canterberry, line 97: This template feels like overkill to me. If every single value is configurable, then it offers little benefit over a user just writing a separate manifest and deploying that. Instead, I'd propose just writing the bare minimum necessary for a functional service monitor in this template, and reduce its configurability to the bare essentials. For fine-grained configurability, the user can disable this chart from deploying a service monitor and the user can deploy their own alongside this chart rather than through it.

values.yaml
```diff
@@ -109,6 +111,19 @@ configData:
...
+    ## Set values here to override the defaults
+    # debug:
+      ## The address and port the debug server should listen on
+      # addr: 0.0.0.0:5001
+      ## Prometheus metrics specific configuration
+      # prometheus:
+          ## Do not disable this if you use metrics.enabled: true
+          # enabled: true
+          ## Make sure to adjust metrics.serviceMonitor.path accordingly
+          # path: /metrics
```
canterberry, line 126: Just thinking aloud and paraphrasing to make sure I understand how this works:

 * To enable metrics, all a user has to do is set `metrics.enabled: true` in the values provided to this chart. If that is set, then a default configuration (as described above) will be injected into the registry's configuration to activate the debug interface and expose it on a default port.
 * If `metrics.enabled: true` is set, and `configData` already has a debug interface configuration, then this chart will derive the relevant port/path/etc from that registry config.

If those assumptions are correct, that seems like reasonable behavior to me.

values.yaml
```diff
@@ -147,3 +162,46 @@ extraVolumes: []
...
+##
+metrics:
+  enabled: true
+  ## The name of the service resources port for the metrics interface
+  servicePortName: debug
+  ## The service port of the services clusterIP
+  servicePort: 5001
+
+  ## The container port name of the registry pods
+  targetPortName: debug
```
canterberry, line 177: I think it's probably okay to be opinionated about what these port names are, rather than making them configurable. Besides, I'm pretty sure the port names are just an implementation detail anyway with no impact outside of how this chart internally "glues" the service/deployment/pod together.

canterberry:
Closing this PR to tidy up the list of ongoing work. Please feel free to reopen or continue the discussion.

PR #7, opened 2020-12-17
Add options for using as a pull through cache

matthewmelvin:
Adds the options needed to run the registry as a pull through cache as per docker's recipe.

https://docs.docker.com/registry/recipes/mirror/


canterberry:
Thanks for the contribution! Looking at the documentation at the link you provided, though, it looks like this is already supported. The `configData:` property allows a configuration file to be provided which may include the `proxy:` property -- much like is proposed in the changeset here -- to configure a pull-through cache. This is a common use case, I imagine! I'm sure an example config for a pull-through cache would be a welcome addition to the README.

canterberry:
Closing this PR to tidy up the list of ongoing work. However, please feel free to reopen or continue the discussion.

PR #8, opened 2021-01-26
Rename port to match Istio naming conventions

nightscape:
Istio requires the service port to be named like `<protocol>[-<suffix>]`.
If possible, it would be good to rename it here to match this requirement.

### DIFF
--- a/templates/service.yaml
+++ b/templates/service.yaml
@@ -32,7 +32,7 @@ spec:
   ports:
     - port: {{ .Values.service.port }}
       protocol: TCP
-      name: {{ .Values.service.name }}
+      name: http-{{ .Values.service.name }}
       targetPort: 5000
 {{- if (and (eq .Values.service.type "NodePort") (not (empty .Values.service.nodePort))) }}
       nodePort: {{ .Values.service.nodePort }}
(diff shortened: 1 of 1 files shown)


templates/service.yaml
```diff
@@ -32,7 +32,7 @@ spec:
   ports:
     - port: {{ .Values.service.port }}
       protocol: TCP
-      name: {{ .Values.service.name }}
+      name: http-{{ .Values.service.name }}
```
canterberry, line 35: Is there a different prefix for `https`? The registry can be configured to serve via HTTP or HTTPS, so if there is a different prefix for https, we'll want to take that into account here.

canterberry:
Thanks for the contribution, @nightscape! One small comment on the protocol name to use. Rationale for the proposed change seems perfectly reasonable and sensible, and I don't see where it would risk breaking any existing configurations.

templates/service.yaml
nightscape: I hopefully fixed this by checking if `{{ .Values.tlsSecret }}` is set.

nightscape:
Hi @canterberry, I tested the changes locally both with and without a `tlsSecretName` and both versions work correctly, i.e. set the port name to `http-5000` or `https-5000` respectively.
From my side this is good to merge.

templates/service.yaml
canterberry: That seems reasonable to me. :+1:

canterberry (approved):
:+1: LGTM

PR #9, opened 2021-02-02
Allow specifying extra environment variables

vixus0:
The registry image has a bunch of configuration provided through environment variables that isn't covered by the chart values.
It doesn't really make sense to try and cover everything through values, so we should allow users to specify additional environment variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ their default values.
 | `ingress.tls`               | Ingress TLS configuration (YAML)                                                           | `[]`            |
 | `extraVolumeMounts`         | Additional volumeMounts to the registry container                                          | `[]`            |
 | `extraVolumes`              | Additional volumes to the pod                                                              | `[]`            |
+| `extraEnv`                  | Additional environment variables for the registry container                                | `[]`            |
 
 Specify each parameter using the `--set key=value[,key=value]` argument to
 `helm install`.
--- a/templates/deployment.yaml
+++ b/templates/deployment.yaml
@@ -158,6 +158,9 @@ spec:
             - name: REGISTRY_STORAGE_DELETE_ENABLED
               value: "true"
 {{- end }}
+{{- if .Values.extraEnv }}
+{{ toYaml .Values.extraEnv | indent 12 }}
+{{- end }}
           volumeMounts:
 {{- if .Values.secrets.htpasswd }}
             - name: auth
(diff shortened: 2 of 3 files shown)


canterberry (commented):
:+1: LGTM. I don't see any obvious negative side effects of this change that would impact existing installations. However, some questions I'd like to answer first to give more confidence:

 1. Is there a specific use case this PR is designed to solve? I'd like to better understand **why** a user might need or want to specify an additional set of arbitrary environment variables.
 2. What would happen if defining extra environment variables that conflict with ones set explicitly by this chart?
 2. Are duplicate names allowed?
 3. Do names declared later override the values of names declared earlier?
 4. If so, is it desirable for values set via extraEnvs to override any or all of the ones hard-coded, derived, or otherwise set?

jwitko:
@canterberry As an answer to #1, this chart is currently unusable for my use-case (so I have to pull it locally and modify it) because its missing this feature.  My use-case is simple, I have an environment that is completely behind proxies in terms of talking to the internet.  This feature would enable me to set things like `http_proxy` environment variable to work around that.

canterberry:
Thanks @jwitko for the context. It's helpful to understand the use case.

Starting with [Docker Engine 17.07](https://docs.docker.com/engine/release-notes/17.07/) (release August 2017), there's a way to [universally configure proxies for Docker to use](https://docs.docker.com/network/proxy/).
For Kubernetes 1.19+, there's a [configuration option to set a `proxy-url` in a kubeconfig context](https://github.com/kubernetes/client-go/issues/351#issuecomment-626752837).

I can see that neither of those would be helpful here, and the `http_proxy` and `https_proxy` environment variables are the way to go.

To that end, I do think that adding support for injecting an arbitrary set of environment variables is a bit of a sledgehammer solution to this problem! Rather, I suggest a more targeted solution: adding `proxy.http` and `proxy.https` values, which if present, get sprinkled into the container's environment variables `HTTP_PROXY` and/or `HTTPS_PROXY` if present.

artlogic:
@canterberry I actually came here looking for this very feature. I have a use case that requires that I disable or customize [manifest validation](https://github.com/distribution/distribution/blob/main/docs/configuration.md#validation). At the moment I'm looking at using a slightly customized version of this chart because of this.

Would you suggest continuing work on this PR, adding a PR for that particular group of settings or some other approach?

canterberry:
@artlogic It looks manifest validation can be configured via the [registry configuration file](https://docs.docker.com/registry/configuration/#list-of-configuration-options) which should be possible to override via the [`configData`](https://github.com/twuni/docker-registry.helm/#configuration) values (i.e: `configData.validation.manifests.urls.allow` and `configData.validation.manifests.urls.deny`). Would that work for your use case?

PR #11, opened 2021-02-07
Add SeviceAccount and Secrets

ichasco-heytrade:
With these changes a ServiceAccount can be created and create Secrets rendered


templates/deployment.yaml
```diff
@@ -27,12 +27,22 @@ spec:
...
 {{- end }}
     spec:
-      {{- if .Values.imagePullSecrets }}
+    {{- if .Values.serviceAccount.create }}
+    {{- if .Values.serviceAccount.name }}
+      serviceAccountName: {{ .Values.serviceAccount.name }}
+    {{- else }}
+      serviceAccountName: {{ include "docker-registry.fullname" . }}
+    {{- end }}
+    {{- end }}
```
canterberry, line 44: When `serviceAccount.create` is false, but `serviceAccount.name` is set, we still want to set `serviceAccountName` here. Only if `create: false` and `name` is empty/unset do we want to abstain from setting `serviceAccountName` here.

canterberry (changes requested):
The portion of this PR for adding a service account looks good, although I spotted one minor defect (comment inline).

Could you remove the bit related to `envRenderSecret` from this PR? There is another PR open related to adding additional environment variables, and I think that's something that could benefit from separate review, as it creates another potentially conflicting pattern for configuring this chart and I'd like to be able to merge this PR for the service account while that discussion continues.

SimonRupar:
@ichasco-heytrade is it possible that you update this PR? This feature would benefit me greatly. With your permission I can also update PR according to @canterberry.

ichasco-heytrade:
Done, can you confirm that it's everything ok? 

Thanks

values.yaml
```diff
@@ -12,6 +12,12 @@ updateStrategy: {}
 podAnnotations: {}
 podLabels: {}
 
+serviceAccount:
+  create: false
+  # name:
+  ## Annotations for the Service Account
```
canterberry, line 18: This comment can be removed.

values.yaml
```diff
@@ -12,6 +12,12 @@ updateStrategy: {}
 podAnnotations: {}
 podLabels: {}
 
+serviceAccount:
```
canterberry, line 15: The README should be updated to describe these new values.

canterberry (changes requested):

templates/deployment.yaml
canterberry: This comment is still relevant and has not been addressed.

SimonRupar:
@ichasco-heytrade, would you be able to adjust to requested change in MR?

templates/deployment.yaml
canterberry: Can you explain how you addressed this feedback? The discussion was marked as resolved, but the code does not appear to have changed and there is no reply to my comment.

SimonRupar: @ichasco-heytrade, will you be able to update this PR shortly?

templates/deployment.yaml
```diff
@@ -28,11 +28,18 @@ spec:
...
-{{- if $.Values.podAnnotations }}
-{{ toYaml $.Values.podAnnotations | indent 8 }}
+{{- if .Values.podAnnotations }}
+{{ toYaml .Values.podAnnotations | indent 8 }}
 {{- end }}
     spec:
-      {{- if .Values.imagePullSecrets }}
+    {{- if .Values.serviceAccount.create }}
+    {{- if .Values.serviceAccount.name }}
+      serviceAccountName: {{ .Values.serviceAccount.name }}
```
SimonRupar, line 37: This causes issues because of whitelines. 
```
unable to parse YAML: error converting YAML to JSON: yaml: line 25: did not find expected key
```

canterberry:
Incorporated into https://github.com/twuni/docker-registry.helm/pull/44 and shipped in v1.16.0.

PR #15, opened 2021-04-25
Add support for service.externalIPs

drewhemm:
This PR adds the ability to define externalIPs for the service, which is useful for situations where load balancers are not available. 

Our use case is a single node cluster, which with an externalIP enables us to pull and push images without having to resort to using a nodePort, which does not work well with rolling updates.

### DIFF
--- a/templates/service.yaml
+++ b/templates/service.yaml
@@ -16,6 +16,12 @@ spec:
 {{- if (and (eq .Values.service.type "ClusterIP") (not (empty .Values.service.clusterIP))) }}
   clusterIP: {{ .Values.service.clusterIP }}
 {{- end }}
+{{- if .Values.service.externalIPs }}
+  externalIPs:
+    {{- range .Values.service.externalIPs }}
+    - {{ . | quote }}
+    {{- end }}
+{{- end }}
 {{- if (and (eq .Values.service.type "LoadBalancer") (not (empty .Values.service.loadBalancerIP))) }}
   loadBalancerIP: {{ .Values.service.loadBalancerIP }}
 {{- end }}
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ their default values.
 | `service.port`              | TCP port on which the service is exposed                                                   | `5000`          |
 | `service.type`              | service type                                                                               | `ClusterIP`     |
 | `service.clusterIP`         | if `service.type` is `ClusterIP` and this is non-empty, sets the cluster IP of the service | `nil`           |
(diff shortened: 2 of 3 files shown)


templates/service.yaml
```diff
@@ -16,6 +16,12 @@ spec:
 {{- if (and (eq .Values.service.type "ClusterIP") (not (empty .Values.service.clusterIP))) }}
   clusterIP: {{ .Values.service.clusterIP }}
 {{- end }}
+{{- if (not (empty .Values.service.externalIPs)) }}
+  externalIPs:
+    {{- range .Values.service.externalIPs }}
+    - {{ . | quote }}
+    {{- end }}
+{{- end }}
```
canterberry, line 24: This block (L19-L24) could be simplified using a combination of helpers (`toYaml` and `nindent`) and collapsing the if-conditional expression (`if not empty X` can be reduced to `if X`):

```yaml
{{- if .Values.service.externalIPs }}
  externalIPs: {{ toYaml .Values.service.externalIPs | nindent 4 }}
{{- end }}
```

Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart for Docker Registry
 name: docker-registry
-version: 1.10.1
+version: 1.11.0
```
canterberry, line 4: Generally, it's a best practice to avoid tethering a version increment to a PR's changeset.

Versioning is a responsibility of the release team, and there may be other PRs going on in parallel that may otherwise have conflicting version increments. It's helpful for any project, though, to *suggest* whether a changeset warrants a patch, minor, or major version increment in the PR's description. Manual version increments could conflict, too, in projects with automated releases, like [semantic release](https://github.com/semantic-release/semantic-release).

canterberry (changes requested):
Initial review complete. Thanks for the contribution!

Overall, the changeset looks good. Just a couple of bits of minor feedback, inline.

The PR's title and commit message effectively describe the changeset, but I would have liked to see the PR description convey **why** (e.g: what motivated you to propose the change? what benefits does the changeset provide?), even if it may seem obvious to you or me. PRs with empty descriptions or context are less likely to receive attention or consideration by maintainers or the community, so a little elaboration can go a long way!

Thank you! Once you push up the small tweaks (remove version increment, replace template block, add a PR description), we're good to go!

canterberry (approved):
:+1: LGTM!

SimonRupar:
@drewhemm, I see that this MR is already approved, but needs signed git commits. Can you please sign your commits?

canterberry:
Closing this PR, as it seems to have been abandoned without having addressed the feedback given.

PR #18, opened 2021-06-11
Add support for registry proxy/mirror

jrhorner1:
This pull request adds support for configuring the registry as a proxy or pull-through cache, as documented [here](https://docs.docker.com/registry/recipes/mirror/). 

My use case for this configuration is to avoid [pull rate limits](https://docs.docker.com/docker-hub/download-rate-limit/) in kubernetes clusters where one may be deploying multiple applications at once such as after a cluster is initially created or when testing different configurations of an application, requiring multiple deployments.


canterberry (dismissed):
:+1: LGTM!

jrhorner1:
Got the commits signed (with the correct email this time).

canterberry (approved):

PR #20, opened 2021-08-02
Add support for having external proxy secrets

joaosa:
When using GitOps tools like Flux, we don't want the contents of values.yaml to be added to Git.
 As such, it is desirable to load them from secrets created a priori.

 This PR adds support for this, on the proxy config, through a `authExternalSecret` key that refs a specific secret in order to fulfill this usecase.

 Let me know if this makes sense :)

### DIFF
--- a/templates/deployment.yaml
+++ b/templates/deployment.yaml
@@ -160,12 +160,12 @@ spec:
             - name: REGISTRY_PROXY_USERNAME
               valueFrom:
                 secretKeyRef:
-                  name: {{ template "docker-registry.fullname" . }}-secret
+                  name: {{ if .Values.proxy.secretRef }}{{ .Values.proxy.secretRef }}{{ else }}{{ template "docker-registry.fullname" . }}-secret{{ end }}
                   key: proxyUsername
             - name: REGISTRY_PROXY_PASSWORD
               valueFrom:
                 secretKeyRef:
-                  name: {{ template "docker-registry.fullname" . }}-secret
+                  name: {{ if .Values.proxy.secretRef }}{{ .Values.proxy.secretRef }}{{ else }}{{ template "docker-registry.fullname" . }}-secret{{ end }}
                   key: proxyPassword
 {{- end }}
 {{- if .Values.persistence.deleteEnabled }}
--- a/values.yaml
+++ b/values.yaml
@@ -103,6 +103,9 @@ proxy:
   remoteurl: https://registry-1.docker.io
   username: ""
   password: ""
+  # the ref for a secret stored outside of this chart
+  # Keys: proxyUsername, proxyPassword
+  secretRef: ""
 
 configData:
   version: 0.1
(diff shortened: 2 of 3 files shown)


canterberry (changes requested):
Thanks!

Externalizing sensitive values into secrets is certainly a good idea. There are lots of good reasons to do so, so :+1: on the premise.

The strategy for externalizing the registry proxy username and password makes sense, and jives with the current methodology.

My only request is to update the README to document this new value, and indicate which keys are expected to be present in the referenced secret. I'd also propose, for clarity, naming the variable `proxy.secretRef`, to clearly indicate it as a reference to a K8s resource. I don't feel strongly about the `auth` prefix, but I do think it's a little redundant and constricting.

joaosa:
> Thanks!
> 
> Externalizing sensitive values into secrets is certainly a good idea. There are lots of good reasons to do so, so 👍 on the premise.
> 
> The strategy for externalizing the registry proxy username and password makes sense, and jives with the current methodology.
> 
> My only request is to update the README to document this new value, and indicate which keys are expected to be present in the referenced secret. I'd also propose, for clarity, naming the variable `proxy.secretRef`, to clearly indicate it as a reference to a K8s resource. I don't feel strongly about the `auth` prefix, but I do think it's a little redundant and constricting.

Thank you. That makes a lot of sense. I updated the PR to match your feedback 👍 . Let me know if it this is what you had in mind

values.yaml
```diff
@@ -103,6 +103,9 @@ proxy:
   remoteurl: https://registry-1.docker.io
   username: ""
   password: ""
+  # the ref for a secret stored outside of this chart
+  # proxyUsername and proxyPassword are the keys to include in the said secret
```
canterberry, line 107: A very minor suggestion to reduce verbosity on this line:

```
# Keys: proxyUsername, proxyPassword
```

canterberry (dismissed):
:+1: LGTM!

Small, targeted changeset. Well articulated value proposition. Sensible strategy. Backwards compatible.

values.yaml
joaosa: Alright updated :)

canterberry (approved):

canterberry (approved):

templates/deployment.yaml
```diff
@@ -160,12 +160,12 @@ spec:
...
               valueFrom:
                 secretKeyRef:
-                  name: {{ template "docker-registry.fullname" . }}-secret
+                  name: {{ if .Values.proxy.secretRef }}{{ .Values.proxy.secretRef }}{{ else }}{{ template "docker-registry.fullname" . }}-secret{{ end }}
                   key: proxyUsername
             - name: REGISTRY_PROXY_PASSWORD
               valueFrom:
                 secretKeyRef:
-                  name: {{ template "docker-registry.fullname" . }}-secret
+                  name: {{ if .Values.proxy.secretRef }}{{ .Values.proxy.secretRef }}{{ else }}{{ template "docker-registry.fullname" . }}-secret{{ end }}
```
lpakula, line 168: I would question that appending _-secret_ to the **secretRef** is a good idea. It's confusing and not documtented

lpakula:
When is the plan to make an official release with this change?

Thanks

canterberry:
These changes are included in v1.12.0, released moments ago!

PR #25, opened 2021-08-26
Add keys proxyUsername and proxyPassword with empty string to docker-registry-secret if values are not set

pavankumar-go:
Setting up docker-registry proxy with no password & username created the secret without `proxyUsername` and `proxyPassword` fields, causing pod to fail if `proxy.enabled = true` with error 
```
error: "couldn't find key proxyUsername in Secret <namespace>/docker-registry-secret"
```

### DIFF
--- a/templates/secret.yaml
+++ b/templates/secret.yaml
@@ -38,7 +38,11 @@ data:
   
   {{- if .Values.proxy.username }}
   proxyUsername: {{ .Values.proxy.username | b64enc | quote }}
+  {{- else }}
+  proxyUsername: ""
   {{- end }}
   {{- if .Values.proxy.password }}
   proxyPassword: {{ .Values.proxy.password | b64enc | quote }}
+  {{- else }}
+  proxyPassword: ""
   {{- end }}
--- a/templates/secret.yaml
+++ b/templates/secret.yaml
@@ -35,14 +35,5 @@ data:
   swiftPassword: {{ .Values.secrets.swift.password | b64enc | quote }}
     {{- end }}
   {{- end }}
-  
-  {{- if .Values.proxy.username }}
-  proxyUsername: {{ .Values.proxy.username | b64enc | quote }}
-  {{- else }}
-  proxyUsername: ""
-  {{- end }}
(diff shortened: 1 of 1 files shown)


templates/secret.yaml
```diff
@@ -38,7 +38,11 @@ data:
   
   {{- if .Values.proxy.username }}
   proxyUsername: {{ .Values.proxy.username | b64enc | quote }}
+  {{- else }}
+  proxyUsername: ""
   {{- end }}
   {{- if .Values.proxy.password }}
   proxyPassword: {{ .Values.proxy.password | b64enc | quote }}
+  {{- else }}
+  proxyPassword: ""
```
canterberry, line 47: Rather than if-blocks, would a ` | default ""` work here?

e.g: `proxyPassword: {{ .Values.proxy.password | default "" | b64enc | quote }}`

If so, same could be done with proxyUsername and make this changeset super tiny!

canterberry (changes requested):
Goal makes sense, to support an anonymous proxy. Small, targeted changeset. One minor request, if possible, to tidy things up even more.

templates/secret.yaml
pavankumar-go: Yes, i agree.
have pushed the new changes using `default` function.

canterberry (approved):

templates/secret.yaml
pavankumar-go: Hi,
Yeah, i have removed the [If-blocks](https://github.com/twuni/docker-registry.helm/pull/25/files) since that returns false when `.Values.proxy.username = "`.

stpierre:
Could this be merged?

canterberry:
Looks like there are still some failing checks, probably commit signing, but once that's resolved, changeset looksgood to merge.

pavankumar-go:
@canterberry my bad, didn't notice the commit signing. Have pushed by signing previous commits.

canterberry (dismissed):

canterberry (approved):

canterberry:
Released in v1.13.1.

PR #27, opened 2021-09-06
Fix for issue number 19, ingress current api version and previous api version…

karezza:
The current version of kubernetes is unable to install this helm chart.  Update Ingress api version to the current stable version used by kubernetes 1.22.*.  Update the else version to use the previous beta version.

### DIFF
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -1,7 +1,7 @@
 {{- if .Values.ingress.enabled -}}
+{{- $apiVersion := .Capabilities.APIVersions -}}
 {{- $serviceName := include "docker-registry.fullname" . -}}
 {{- $servicePort := .Values.service.port -}}
-{{- $apiVersion := .Capabilities.APIVersions -}}
 {{- $path := .Values.ingress.path -}}
 apiVersion: {{- if $apiVersion.Has "networking.k8s.io/v1" }} networking.k8s.io/v1 {{- else }} networking.k8s.io/v1beta1 {{- end }}
 kind: Ingress
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -20,6 +20,9 @@ metadata:
       {{ $key }}: {{ $value | quote }}
     {{- end }}
 spec:
+{{- if $apiVersion.Has "networking.k8s.io/v1" }}
+  ingressClassName: {{ .Values.ingress.ingressClassName }}
+{{- end }}
   rules:
     {{- range $host := .Values.ingress.hosts }}
     - host: {{ $host }}
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -2,7 +2,7 @@
 {{- $serviceName := include "docker-registry.fullname" . -}}
 {{- $servicePort := .Values.service.port -}}
 {{- $path := .Values.ingress.path -}}
-apiVersion: {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }} networking.k8s.io/v1beta1 {{- else }} extensions/v1beta1 {{- end }}
+apiVersion: {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }} networking.k8s.io/v1 {{- else }} networking.k8s.io/v1beta1 {{- end }}
 kind: Ingress
 metadata:
   name: {{ template "docker-registry.fullname" . }}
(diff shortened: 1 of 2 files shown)


tarioch:
Careful, there is more required than just changing the version, see https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122

vivekpatani:
@karezza 

```
{{- if .Values.ingress.enabled -}}
{{- $serviceName := include "docker-registry.fullname" . -}}
{{- $servicePort := .Values.service.port -}}
{{- $path := .Values.ingress.path -}}
apiVersion: {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }} networking.k8s.io/v1 {{- else }} extensions/v1 {{- end }}
kind: Ingress
metadata:
  name: {{ template "docker-registry.fullname" . }}
  labels:
    app: {{ template "docker-registry.name" . }}
    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
    release: {{ .Release.Name }}
    heritage: {{ .Release.Service }}
{{- if .Values.ingress.labels }}
{{ toYaml .Values.ingress.labels | indent 4 }}
{{- end }}
  annotations:
    {{- range $key, $value := .Values.ingress.annotations }}
      {{ $key }}: {{ $value | quote }}
    {{- end }}
spec:
  rules:
    {{- range $host := .Values.ingress.hosts }}
    - host: {{ $host }}
      http:
        paths:
          - path: {{ $path }}
            pathType: Prefix
            backend:
              service:
                name: {{ $serviceName }}
                port:
                  number: {{ $servicePort }}
    {{- end -}}
  {{- if .Values.ingress.tls }}
  tls:
{{ toYaml .Values.ingress.tls | indent 4 }}
  {{- end -}}
{{- end -}}
```

The `service` section towards the end is broken up into multiple entities too, and `PathType: <ImplementationSpecific>` is a neccesary field. I'm not sure if these are the only changes needed, but this worked for me.

canterberry (approved):
Assuming that `extensions/v1beta1` is EOL, which I think it effectively is at this point, I don't mind breaking support for it.

kervel:
i think https://github.com/Kapernikov/docker-registry.helm/commit/22eca8f7680c1289dd094ffe36f44fbbedde14f5 this is what is needed ...

karezza:
Added 'ingressClassName' and also defined a default value in values.yaml, it is my understanding that although an ingress can be created without this with a default being automatically selected, instead people should always have it defined in their ingress going forward and not in the annotations (where this was sometimes happening before, but not in a clearly defined way).

kervel:
i went  thru the kubernetes history to see how long networking.k8s.io/v1 is already there, and apparently its only there since 1.19 (which is 1 year ago). actually astonished that kubernetes breaks compat with old API only 1 year after releasing the new one...

so i guess it makes sense to try to also support the v1beta1.

ElArtista:
I suppose #33 is relevant to this?

DBarthe:
Hello !

Could we just add :
```
apiVersion: {{- if $apiVersion.Has "networking.k8s.io/v1" }} networking.k8s.io/v1 {{- else if $apiVersion.Has "networking.k8s.io/v1beta1" }} networking.k8s.io/v1beta1 {{- else }} extensions/v1beta1 {{- end }}
```
So that we can support 1.22 while not breaking any older version ?

ElArtista (approved):

templates/ingress.yaml
```diff
@@ -1,8 +1,9 @@
 {{- if .Values.ingress.enabled -}}
+{{- $apiVersion := .Capabilities.APIVersions -}}
```
canterberry, line 2: Should be `$apiVersions`

values.yaml
```diff
@@ -32,6 +32,7 @@ service:
   # foo.io/bar: "true"
 ingress:
   enabled: false
+  ingressClassName: nginx
```
canterberry, line 35: `ingress.ingressClassName` is redundant -- instead, should be `ingress.className`.

Please add documentation for this value to the README as well.

canterberry (changes requested):
Thank you for this PR. I have a couple of very small requests, but aside from that this changeset looks good. In addition to looking over the changeset, I ran through the exercise of making the same set of changes by cross-referencing against the `kubectl explain ingress` output on a modern cluster. Line-for-line, I came to pretty much the same result.

canterberry:
Superceded by #48.

PR #29, opened 2021-09-29
Add support for S3 bucket data prefix

SimonRupar:
Add support for S3 bucket to prefix all data

https://docs.docker.com/registry/storage-drivers/s3/
> This is a prefix that is applied to all S3 keys to allow you to segment data in your bucket if necessary.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -91,6 +91,7 @@ secrets:
 #  region: us-east-1
 #  regionEndpoint: s3.us-east-1.amazonaws.com
 #  bucket: my-bucket
+#  rootdirectory: /object/prefix
 #  encrypt: false
 #  secure: true
 
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ their default values.
 | `s3.region`                 | S3 region                                                                                  | `nil`           |
 | `s3.regionEndpoint`         | S3 region endpoint                                                                         | `nil`           |
 | `s3.bucket`                 | S3 bucket name                                                                             | `nil`           |
+| `s3.rootdirectory`          | S3 prefix that is applied to allow you to segment data                                     | `nil`           |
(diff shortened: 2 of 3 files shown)


values.yaml
```diff
@@ -91,6 +91,7 @@ secrets:
 #  region: us-east-1
 #  regionEndpoint: s3.us-east-1.amazonaws.com
 #  bucket: my-bucket
+#  rootdirectory: my-docker
```
canterberry, line 94: It would be helpful for the example to illustrate this value as a path, to help the reader understand where slashes should be.

Which of the following is correct if I want object `foo` to become `object/prefix/foo`?

 • `object/prefix`
 • `/object/prefix`
 • `object/prefix/`
 • `/object/prefix/`

`my-docker` is not descriptive enough to convey confidence that it is conveying a path. Including a slash in the middle helps, and at the beginning and/or end as necessary. Having the example value itself convey through the words it uses that it is, unambiguously, an object prefix, further empowers the reader to identify the correct form for their use case.

canterberry (changes requested):
Looks good, with one minor change for the example value. Straightforward, well-scoped changeset.

values.yaml
SimonRupar: Hello @canterberry, thank you very much for quick reply. 
I have tested with all the cases you present, and the propagation works in all the cases you present. I agree with update of example so I have set to ```/object/prefix``` because it seems to me, it may be more descriptive.

canterberry (approved):

SimonRupar:
@canterberry, sorry I didn't noticed that my signing key was not ok. Can you confirm that now everything is ok?

canterberry (approved):

canterberry:
Released just now in v1.13.2.

SimonRupar:
@canterberry, great thank you!

PR #32, opened 2021-10-01
Add minimal config to enable metrics

vonera:
- The default configData configuration enables the debug server but does not expose it. 
- The user can enable metrics by setting configData.http.debug.prometheus.enabled:=true
- If metrics is enabled the metrics port (default 5001) gets exposed by the pod. 
- The service and service-monitor configuration is left to the user (for simplicity and security)

### DIFF
--- a/templates/deployment.yaml
+++ b/templates/deployment.yaml
@@ -54,6 +54,11 @@ spec:
           - /etc/docker/registry/config.yml
           ports:
             - containerPort: 5000
+{{- if .Values.configData.http.debug.prometheus.enabled }}
+            - containerPort: {{ (split ":" .Values.configData.http.debug.addr)._1 }}
+              name: metrics
+              protocol: TCP
+{{- end }}
           livenessProbe:
             httpGet:
 {{- if .Values.tlsSecretName }}
--- a/values.yaml
+++ b/values.yaml
@@ -122,8 +122,6 @@ configData:
     addr: :5000
     headers:
       X-Content-Type-Options: [nosniff]
-    ## https://docs.docker.com/registry/configuration/#debug
-    ## to enable metrics set debug.prometheus.enabled true
     debug:
       addr: :5001
       prometheus:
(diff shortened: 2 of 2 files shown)


SimonRupar:
+1 for this feature

templates/deployment.yaml
```diff
@@ -54,6 +54,11 @@ spec:
           - /etc/docker/registry/config.yml
           ports:
             - containerPort: 5000
+{{- if .Values.configData.http.debug.prometheus.enabled }}
```
canterberry, line 57: This is a deeply nested configuration path, increasing the risk that some ancestor along the way is nil. Is there a "maybe" syntax or helper function that can be used here to short-circuit to false in that case?

templates/deployment.yaml
```diff
@@ -54,6 +54,11 @@ spec:
...
+            - containerPort: {{ (split ":" .Values.configData.http.debug.addr)._1 }}
```
canterberry, line 58: Is the debug addr guaranteed to be present? On mobile, so I can't see the whole diff easily to determine whether there's an existing guard clause around this, but since it's referencing something that wasn't being referenced before, I wanted to double check.

values.yaml
```diff
@@ -122,6 +122,13 @@ configData:
     addr: :5000
     headers:
       X-Content-Type-Options: [nosniff]
+    ## https://docs.docker.com/registry/configuration/#debug
+    ## to enable metrics set debug.prometheus.enabled true
```
canterberry, line 126: These comments aren't really necessary, and could easily drift due to url changes or config path changes later on. OK to remove the above two comment lines.

canterberry (changes requested):
Seems fine, just need some more defensive checks. Any new configuration values, please also update the table in the README to help folks discover them more easily.

values.yaml
vonera: Sure, I will fix that

templates/deployment.yaml
vonera: he can continue, I close this

PR #35, opened 2021-10-11
Support additional env variables

vvanouytsel:
This commits adds support to set extra ENV variables.
Much in the same way as `extraVolumeMounts` and `extraVolumes`.


canterberry (changes requested):
Changeset looks good. Can you update the table in the README to include the new configuration value?

PR #37, opened 2021-10-14
Enable metrics via Prometheus Operator

SimonRupar:
Added support to enable metrics of docker-registry and enabled integration with Prometheus Operator (https://github.com/prometheus-operator/prometheus-operator)

This enables metrics and native seamless integration of docker-registry setup with Prometheus Operator. This functionality enables both passing of metrics and alert configuration to Prometheus. This functionality is by default disable and user can enable it.

Specification if metrics are enabled:
- Metrics are be exposed on Pod and Service. 
- Pod port for metrics is set via `configData` while port for Service is set via values ( `.Values.metrics.port` )
- ServiceMonitor and PrometheusRule are deployed which are part of Prometheus Operator `monitoring.coreos.com/v1`

Credits for exposing metrics: https://github.com/twuni/docker-registry.helm/pull/32

Example of configuration (Values.yaml) with enabled metrics and Alerts would look like:

```
....
metrics:
  enabled: true
  serviceMonitor:
    enabled: true
    labels:
      prometheus: prometheus
  prometheusRule:
    enabled: true
    labels :
      prometheus: prometheus
    rules:
      - alert: TooMany400s
        annotations:
          description: Too many 4XXs
          summary: More than 5% of the all requests did return 4XX, this requires your
            attention
        expr: 100 * ( sum( registry_http_requests_total{code=~"4.+"} ) / sum(registry_http_requests_total)
          ) > 5
        for: 1m
        labels:
          severity: critical 
....
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ their default values.
 | `ingress.path`              | Ingress service path                                                                       | `/`             |
 | `ingress.hosts`             | Ingress hostnames                                                                          | `[]`            |
 | `ingress.tls`               | Ingress TLS configuration (YAML)                                                           | `[]`            |
-| `metrics.enable`            | Enable metrics on Service                                                                  | `false`         |
+| `metrics.enabled`           | Enable metrics on Service                                                                  | `false`         |
 | `metrics.port`              | TCP port on which the service metrics is exposed                                           | `5001`          |
 | `metrics.serviceMonitor.annotations` | Prometheus Operator ServiceMonitor annotations                                    | `{}`            |
 | `metrics.serviceMonitor.enable` | If true, Prometheus Operator ServiceMonitor will be created                            | `false`         |
--- a/README.md
+++ b/README.md
@@ -91,6 +91,15 @@ their default values.
 | `ingress.path`              | Ingress service path                                                                       | `/`             |
 | `ingress.hosts`             | Ingress hostnames                                                                          | `[]`            |
 | `ingress.tls`               | Ingress TLS configuration (YAML)                                                           | `[]`            |
+| `metrics.enable`            | Enable metrics on Service                                                                  | `false`         |
+| `metrics.port`              | TCP port on which the service metrics is exposed                                           | `5001`          |
+| `metrics.serviceMonitor.annotations` | Prometheus Operator ServiceMonitor annotations                                    | `{}`            |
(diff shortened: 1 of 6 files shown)


metroshica:
Any possibility on getting this merged? Would love to have metrics for our docker registry.

values.yaml
```diff
@@ -110,6 +110,19 @@ proxy:
   # Keys: proxyUsername, proxyPassword
   secretRef: ""
 
+metrics:
+  enable: false
```
canterberry, line 114: Should be `enabled`.

canterberry (changes requested):
Minor changes requested (small typos).

canterberry:
This PR does one important thing: provide the ability to expose the registry's debug port. Without that, I can see how monitoring it may be difficult.

Beyond that is convenience, however, expanding on the scope of this chart perhaps a bit too far, because it leans into a solution (Prometheus) for which users have other choices. That's my non-blocking and loosely held opinion.

As long as it doesn't break anything, though, and remains "opt in" functionality, I see no compelling reason not to approve, given at least a few folks have indicated it would be useful.

README.md
```diff
@@ -91,6 +91,15 @@ their default values.
 | `ingress.path`              | Ingress service path                                                                       | `/`             |
 | `ingress.hosts`             | Ingress hostnames                                                                          | `[]`            |
 | `ingress.tls`               | Ingress TLS configuration (YAML)                                                           | `[]`            |
+| `metrics.enable`            | Enable metrics on Service                                                                  | `false`         |
```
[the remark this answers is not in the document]
SimonRupar: Great catch. Sorry about this. Updated to `metrics.enabled`.

values.yaml
SimonRupar: Updated.

SimonRupar:
@canterberry, true including this may lead in favor of Prometheus Operator, but Prometheus was adopted early by CNCF and wildly used. This nicely round up solution for monitoring and simplify deployment for first time users.
Links: https://www.cncf.io/projects/, https://prometheus.io/docs/introduction/overview/
It is also by default disabled in `values.yaml` so you have to extra enable this. 

But still if there are some concerns, I can move Prometheus operator out of scope of this PR and prepare new MR.

canterberry (approved):

canterberry:
Released in v1.15.0.

PR #38, opened 2021-10-26
Added support for pod autoscaling using Horizontal Pod Autoscaler

SimonRupar:
Added Horizontal Pod Autoscaler to enable autoscaling of pods  
https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/

This enables configuration of autoscaling of pods containing docker registry. This functionality is disabled by default. 

Example of configuration:

```
autoscaling:
  enabled: true
  minReplicas: 2
  maxReplicas: 5
  targetCPUUtilizationPercentage: 50
```

### DIFF
--- a/templates/hpa.yaml
+++ b/templates/hpa.yaml
@@ -1,0 +1,42 @@
+{{- if .Values.autoscaling.enabled }}
+{{- $apiVersions := .Capabilities.APIVersions -}}
+{{- if $apiVersions.Has "autoscaling/v2" }}
+apiVersion: autoscaling/v2
+kind: HorizontalPodAutoscaler
+metadata:
+  name: {{ template "docker-registry.fullname" . }}
+  labels:
+    app: {{ template "docker-registry.name" . }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: {{ template "docker-registry.fullname" . }}
+  minReplicas: {{ .Values.autoscaling.minReplicas }}
+  maxReplicas: {{ .Values.autoscaling.maxReplicas }}
+  metrics:
+{{- with .Values.autoscaling.targetCPUUtilizationPercentage }}
+    - type: Resource
+      resource:
+        name: cpu
+        target:
+          type: Utilization
+          averageUtilization: {{ . }}
+{{- end }}
+{{- with .Values.autoscaling.targetMemoryUtilizationPercentage }}
+    - type: Resource
+      resource:
+        name: memory
+        target:
+          type: Utilization
+          averageUtilization: {{ . }}
+{{- end }}
+{{- with .Values.autoscaling.behavior }}
+  behavior:
+    {{- toYaml . | nindent 4 }}
+{{- end }}
+{{- end }}
+{{- end }}
--- a/templates/hpaV1.yaml
+++ b/templates/hpaV1.yaml
@@ -1,0 +1,22 @@
+{{- if .Values.autoscaling.enabled }}
+{{- $apiVersions := .Capabilities.APIVersions -}}
+{{- if not ($apiVersions.Has "autoscaling/v2") }}
+apiVersion: autoscaling/v1
+kind: HorizontalPodAutoscaler
+metadata:
+  name: {{ template "docker-registry.fullname" . }}
+  labels:
+    app: {{ template "docker-registry.name" . }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: {{ template "docker-registry.fullname" . }}
+  minReplicas: {{ .Values.autoscaling.minReplicas }}
+  maxReplicas: {{ .Values.autoscaling.maxReplicas }}
+  targetCPUUtilizationPercentage: {{ .Values.autoscaling.targetCPUUtilizationPercentage }}
+{{- end }}
+{{- end }}
--- a/values.yaml
+++ b/values.yaml
@@ -163,6 +163,30 @@ podDisruptionBudget: {}
   # maxUnavailable: 1
   # minAvailable: 2
 
+autoscaling:
+  enabled: false
+  minReplicas: 1
+  maxReplicas: 2
+  targetCPUUtilizationPercentage: 60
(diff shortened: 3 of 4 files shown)


SimonRupar:
@canterberry, can I ask you to take a look into this? Can I somehow help with providing some information?

templates/hpa.yaml
```diff
@@ -0,0 +1,35 @@
+{{- if .Values.autoscaling.enabled }}
+apiVersion: autoscaling/v2beta1
```
canterberry, line 2: HPAs seem to be a rapidly evolving feature, available in several different API groups (v1, v2beta1, v2beta2, v2). We have a similar issue in this chart with the Ingress resource -- to maintain support for a spread of Kubernetes versions, special consideration has to be given to which API version(s) are allowed in the cluster to which the chart is being installed.

In the case of Ingress, only two different API groups need to be supported (historically there are three different API groups for Ingress resources, but one of those has reached its EOL). For HPAs, at least three API versions seem to be concurrently active, so special consideration may be needed for each of `v2`, `v2beta2`, `v2beta1`, and possibly `v1`. That would, I imagine, result in a very messy file here. Maybe not, though.

This changeset seems fine, and in alignment with the general strategy of other aspects of this chart in the manner in which it is designed to be configured. My only concern, really, is compatibility. `v2beta1` is too specific and niche, but if this changeset were to be augmented for compatibility with all non-EOL K8s versions, that's all I'd need for a :+1:.

canterberry:
Hi @SimonRupar . Thanks for your contribution here; I don't audit pending PRs as often as perhaps I should. I apologize for not getting to this sooner, it was completely off my radar until today, even though you @-mentioned me. Not sure how I missed it, but I'm sorry! I appreciate your contribution, I've reviewed it, and it's nicely done. I do have one request (comment inline), out of concern for compatibility since this PR is using the v2beta1 variant of HPA.

 * [autoscaling/v2beta2 HorizontalPodAutoscaler](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#horizontalpodautoscaler-v126): introduced in v1.22, EOL in v1.26
 * [autoscaling/v2beta1 HorizontalPodAutoscaler](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#horizontalpodautoscaler-v125): introduced in v1.21, EOL in v1.25
 * [autoscaling/v1 HorizontalPodAutoscaler](https://kubernetes.io/docs/reference/using-api/deprecation-policy/): **NOT** deprecated in v1.23 despite deprecation policy indicating it should be, EOL unknown
 * [autoscaling/v2 HorizontalPodAutoscaler](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.23.md#horizontalpodautoscaler-v2-graduates-to-ga): introduced in v1.23

Since the `autoscaling/v1` version of the HorizontalPodAutoscaler is not deprecated and is available in all supported Kubernetes versions (as of today: v1.20, v1.21, v1.22, v1.23), that would be the most suitable version for this chart to use. Later, when v1.23 is the oldest supported K8s version, I think upgrading that to the `autoscaling/v2` variant makes sense.

templates/hpa.yaml
SimonRupar: Thank you for detailed feedback. 
I agree. I have just updated PR which allows to switch between version `autoscaling/v1` and `autoscaling/v2` with values. By default version is `autoscaling/v1` if `autoscale.enable` is set to `true`. I intentionally created two files to simplify maintenance and phase out when possible. If you prefer, I can also bundle things in same file because specifications as far as I can see are quite close.
If possible please give me a day or two to test this thoroughly.

templates/hpaV1.yaml
```diff
@@ -0,0 +1,21 @@
+{{- if .Values.autoscaling.enabled }}
+{{- if eq .Values.autoscaling.apiVersion "v1" -}}
```
canterberry, line 2: Did you know there is a `.Capabilities.APIVersions` collection available here that can be used to query the target cluster's API version support? It could alleviate the need for an `autoscaling.apiVersion` value -- opportunisticslly use v2 if supported, else the chart could fallback to v1.

values.yaml
```diff
@@ -139,6 +139,27 @@ podDisruptionBudget: {}
...
   # minAvailable: 2
 
+autoscaling:
+  enabled: false
+  apiVersion: v1
+  minReplicas: 1
+  maxReplicas: 2
+  targetCPUUtilizationPercentage: 75
+  targetMemoryUtilizationPercentage: 75 # available only on version v2
+  behavior: {} # available only on version v2
```
canterberry, line 149: Thanks for this comment to clarify why a user might set a value and see it being ignored.

With the suggested removal of the apiVersion prop, though, the comment would lose the benefit of being in close promixity to the version it is referencing and become ambiguous.

Proposed change to more directly identify the constraints of when the value would take effect: "available only on Kubernetes ≥1.23"
(i.e: when the autoscaling/v2 HPA was introduced)

canterberry (changes requested):
Thanks for the quick turnaround! One small optimization requested and related comment revision, but apart from that, LGTM. Quite nicely done all around.

Kudos: In particular, I prefer what you did here over what this chart's doing for compatibility of the _Ingress_ resource -- your approach of defining the two different kinds of resources separately, vs defining them in one file with feature-detection conditionals piecemeal within, I think has resulted in manifests that are easier to follow. Sure there's some duplication because v1 and v2 do have many similarities, it's easier to grok what each is doing because they're separate.

templates/hpaV1.yaml
SimonRupar: I was in dilemma regarding this. I went with approach to enable user selecting version, giving him possibility to decide when he would like to update _HPA_ independent from kubernetes upgrades. Because using `.Capabilities.APIVersions` basically forces the user to upgrade HPA when updating kubernetes cluster version. But on other hand, these changes are always with high-impacting so some migration activities are always in order. Also the same "migration flow" is for _Ingress_. 

I have removed `autoscaling.apiVersion` variable and use `.Capabilities.APIVersions` for _HPA_ resources. I also renamed _HPA_ file for V2 to `hpa.yaml` so when `autoscaling/v1` is truly deprecated, it makes things easier. As suggested it will use `autoscaling/v2` if available else fallback to `autoscaling/v1` (provided `autoscaling.enabled` is set to `true`)

values.yaml
SimonRupar: Updated, and I also updated values definitions inside `ReadME.md`.

templates/hpa.yaml
SimonRupar: I managed to test chart on kubenetes `v1.21.8` (`autoscaling/v1`) and kubernetes `1.23.1`- minikube (`autoscaling/v2`)  and selection of _HPA_ works for me as intended.

canterberry (approved):
LGTM. Thank you, and nicely done!
I appreciate the context and quick turnaround.

canterberry:
Published in v2.1.0.

PR #40, opened 2021-11-11
Add initContainer support

g-linville:
This commit will give the user the ability to specify initContainers directly in the values file, which will then become part of the deployment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,7 @@ their default values.
 | `extraVolumeMounts`         | Additional volumeMounts to the registry container                                          | `[]`            |
 | `extraVolumes`              | Additional volumes to the pod                                                              | `[]`            |
 | `extraEnvVars`              | Additional environment variables to the pod                                                | `[]`            |
+| `initContainers`            | Init containers to be created in the pod                                                   | `[]`            |
 
 Specify each parameter using the `--set key=value[,key=value]` argument to
 `helm install`.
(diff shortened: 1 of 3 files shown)


weisjohn:
@g-linville I'd love to see this land in the chart. Can you provide some more of the rationale as to why you need this?

g-linville:
> @g-linville I'd love to see this land in the chart. Can you provide some more of the rationale as to why you need this?

Thanks @weisjohn. The reason I need this is to adjust file permissions of the registry directory by using `chown` to change ownership to a different user, since I am running the container as a non-root user.

@twuni if you would be willing to merge this, that would be great! I appreciate this Helm chart quite a lot, since it provides such an easy way to run the Docker Registry inside of k8s.

izorster (dismissed):
This seems reasonable

canterberry (changes requested):
Can you add docs to the table in the README for this new value? Then, LGTM as well.

g-linville:
@canterberry done!

canterberry (dismissed):

g-linville:
@canterberry I had to force-update my branch so I could sign my commits with GPG. Can you re-approve? Sorry for the inconvenience.

canterberry (approved):

weisjohn:
@canterberry do you think you could release this?

PR #42, opened 2021-12-03
Support deployment to a namespace

### LINKED ISSUE #41: Add namespaces to metadata
None of the template files have the namespace specified. This means that non-default namespaces are not respected when using this chart

Proposed fix:

Add this to all template YAML files:

```yaml
metadata:
    namespace: {{ .Release.Namespace }}
```

Happy to raise a PR for this if it would be accepted

MrSimonEmms:
The charts didn't specify `metadata.namespace` anywhere so have added support for this. Have used the `{{ .Release.Namespace }}` variable which defaults to `default` if not set. This makes it a non-intrusive change for anyone not wanting to use it, but a useful change for those who do

Fixed #41

Have tested against my own cluster with these commands:

## Default namespace

```shell
helm upgrade \
  --atomic \
  --cleanup-on-fail \
  --install \
  --wait \
  registry .
```

This deploys `registry` to `default` namespace

## Custom namespace

```shell
helm upgrade \
  --atomic \
  --cleanup-on-fail \
  --create-namespace \
  --install \
  --namespace registry \
  --wait \
  registry .
```

This deploys `registry` to `registry` namespace

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ their default values.
 | `proxy.username`            | Remote registry login username                                                             | `nil`           |
 | `proxy.password`            | Remote registry login password                                                             | `nil`           |
 | `proxy.secretRef`           | The ref for an external secret containing the proxyUsername and proxyPassword keys         | `""`            |
+| `namespace`                 | specify a namespace to install the chart to - defaults to `.Release.Namespace`             | `{{ .Release.Namespace }}` |
 | `nodeSelector`              | node labels for pod assignment                                                             | `{}`            |
 | `affinity`                  | affinity settings                                                                          | `{}`            |
 | `tolerations`               | pod tolerations                                                                            | `[]`            |
--- a/templates/configmap.yaml
+++ b/templates/configmap.yaml
@@ -2,6 +2,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ template "docker-registry.fullname" . }}-config
+  namespace: {{ .Values.namespace | default .Release.Namespace }}
   labels:
     app: {{ template "docker-registry.name" . }}
     chart: {{ .Chart.Name }}-{{ .Chart.Version }}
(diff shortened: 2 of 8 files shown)


canterberry (approved):
To install a Helm chart into a specific namespace, can you not pass `--namespace ...` to the helm install/upgrade command? This should, in theory, install all resources to that namespace, as well as the Helm release metadata. If my understanding is correct, the changeset should be a no-op (i.e: since it reflects the default behavior anyway). If the chart currently just installs resources to the default namespace anyway, though, that's certainly undesirable!

In all the Helm charts I have written, I have done as you have proposed in this changeset, so I am inclined to approve. Regardless of whether my assumption above is correct regarding the default behavior, I do think being explicit about the namespace is an improvement.

However, I've read that having the resource namespace match the release metadata namespace is not always desirable. Helm release secrets do accumulate over time, polluting the release namespace. To split that, you'd have essentially the same changeset as here, but `{{ .Values.namespace | default .Release.Namespace }}`, and introduce an optional `namespace` value that chart users can set if desired. That solution, I think, would take care of all desired namespacing scenarios.

MrSimonEmms:
@canterberry specifying `--namespace` may well work with you're using it with `helm install` or `helm upgrade`, but it doesn't seem to work with `helm template` which is my specific use-case - see [gitpod/installer](https://github.com/gitpod-io/gitpod/tree/main/installer) for more details.

I've not come across what you've said about the resource/release namespace matching not being desirable, but I can see that it could be a problem with a library. I'll make the change you suggested and document it in the `values.yaml`

canterberry (dismissed):

MrSimonEmms:
@canterberry apologies, I hadn't signed the commit - done now

canterberry (approved):

canterberry:
Released in v1.15.0.

PR #47, opened 2022-01-01
Add support for stable ingress version networking.k8s.io/v1

flojon:


flojon:
Duplicate of #27 - Forgot to check before creating pull request

PR #51, opened 2022-01-27
🌐 make protocol selection work with istio

jsievenpiper:
Quick PR that renames the metrics port from `metrics` to `http-metrics` to make scraping work automatically with Istio protocol selection: https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/

We can also avoid renaming the port by means of using `appProtocol` on the Service Port spec, but figured if this was viable it works across a wider set of K8S versions.

### DIFF
--- a/templates/deployment.yaml
+++ b/templates/deployment.yaml
@@ -60,7 +60,7 @@ spec:
             - containerPort: 5000
 {{- if .Values.metrics.enabled }}
             - containerPort: {{ (split ":" .Values.configData.http.debug.addr)._1 }}
-              name: metrics
+              name: http-metrics
               protocol: TCP
 {{- end }}
           livenessProbe:
--- a/templates/service.yaml
+++ b/templates/service.yaml
@@ -41,7 +41,7 @@ spec:
 {{- if .Values.metrics.enabled }}
     - port: {{ .Values.metrics.port }}
       protocol: TCP
-      name: metrics
+      name: http-metrics
       targetPort: {{ (split ":" .Values.configData.http.debug.addr)._1 }}
 {{- end }}
   selector:
--- a/templates/servicemonitor.yaml
+++ b/templates/servicemonitor.yaml
@@ -16,6 +16,6 @@ spec:
       release: {{ .Release.Name }}
       heritage: {{ .Release.Service }}
   endpoints:
-  - port: metrics
+  - port: http-metrics
     interval: 15s
 {{- end }}


canterberry (commented):
Thanks for your contribution! It's folks like you that keep the chart relevant and active.

To me, the changeset seems innocuous because it's changing what I perceive to be an arbitrary port name (opaque to k8s itself) from one value to another. For the same reason, I feel like there's more I need to understand about *why* this has any impact, on Istio or otherwise, and what implications that might have for scenarios other than the one you're solving for (i.e: is `metrics` a magic name required for some other important thing to work?).

Could you explain like I'm 5, the problem this changeset aims to remediate, and how it does that? Once I feel more confident that I understand what's going on, I'll feel comfortable approving.

Changeset's fine; just want to make sure I grok the context.

jsievenpiper:
Hey @canterberry! Thanks for the kind words and quick review 🙂 

Yeah, I agree it's not immediately intuitive at all (and as much as I'd love to switch to linkerd, istio is the only one that checks all the boxes for me right now). There may be some gaps in my understanding, but this is how I've reasoned about it in my head. The tl;dr of istio is that it provides mTLS between services by injecting sidecar proxies in each pod which *usually* transparently encrypt/decrypt traffic along with handling authorization between services.  So if `pod-a` wants to make an http request to `pod-b`, they'd just make a plain ol' http request, and by the magic of istio, that'll get upgraded in transit to something that is not only encrypted but that both sides can mutually verify identify. `pod-b` on the other end would listen as if it were plaintext traffic and is none the wiser.

Prometheus itself has a little gotcha in that the way it scrapes metrics _I believe_ because it decides to directly connect to underlying pods via their IP (I assume it does this because there's no reliable hostname for any given individual pod, so an IP is the only/easy way to go) and tools like linkerd and istio need something to tie identity to at the _service_ level with some sort of name that a certificate can be signed for, which may map to one or more pods in a replicated flow.

So when running docker-registry in a strictly-configured service mesh, the metrics endpoint is guarded by mTLS. That's okay, because we can configure Prometheus to scrape using HTTPS and give it a cert/key to verify the identity with. Istio [documents](https://istio.io/latest/docs/ops/integrations/prometheus/#tls-settings) this somewhat poorly, and the [ServiceMonitor](https://prometheus-operator.dev/docs/operator/api/#endpoint) API has space for us to configure it.

This last part is the one I don't have a really great grasp on, but when istio can't determine the underlying protocol type, it's supposed to default to plain TCP, which should be perfectly functional, but something causes the handshake to die unless you identify the traffic as HTTP. My _guess_ is that because you only have one side of the TLS handshake going on (Prometheus is issuing an HTTPS request, which is inherently single-sided TLS instead of mutual), the generic upgrade from TCP -> mTLS fails since only one party is upholding their side of the bargain.

...

..I think!

Anyway! Sorry that was really long-winded. I've noticed that most charts/k8s-deployments that try to be service-mesh friendly tend to just abide by that naming scheme of `protocol[-name]` and that helps at least Istio be chill about the whole thing. 

To directly answer your question: nah, `metrics` isn't special, the lack of `http` as a prefix is what is special. As far as implications: I _think_ it'd be pretty transparent unless someone has deployed this chart and is extending it with another deployment, custom service (maybe a headless service for service discovery with a replicated registry?), in which case it'd be a breaking change if you're considering that part of the API!

Thanks for your consideration!

canterberry (approved):
Thanks for the detailed write-up! Lots happening outside of the chart's domain of responsibility that's helpful to know about, and distilled in a way that makes sense of it all.

Based on what you described, it sounds like istio might be relying on SNI or hostname validation, and with an unknown next protocol, wouldn't know how to inject/sniff the hostname. That's my guess, anyway, for how *http-metrics* would help: by giving it a Host header to look at.

With the significance of the port name demystified, I now feel much better equipped to approve with confidence. Thank you!

PR #56, opened 2022-02-23
Lookup randomly generated haSharedSecret

jthurner:
If haSharedSecret is not defined an existing value on the running
cluster will be reused before generating a random value, preventing
deployments from needlessly rolling on upgrade. The secret can still be
explicitly rotated with "--set secrets.haSharedSecret".

### DIFF
--- a/templates/secret.yaml
+++ b/templates/secret.yaml
@@ -16,9 +16,14 @@ data:
   {{- if .Values.secrets.haSharedSecret }}
   haSharedSecret: {{ .Values.secrets.haSharedSecret | b64enc | quote }}
   {{- else }}
+  {{- $secret := lookup "v1" "Secret" .Release.Namespace (printf "%s-secret" (include "docker-registry.fullname" .)) }}
+    {{- if (dig "data" "haSharedSecret" "" $secret) }}
+  haSharedSecret: {{ $secret.data.haSharedSecret | quote }}
+    {{- else }}
   haSharedSecret: {{ randAlphaNum 16 | b64enc | quote }}
+    {{- end }}
   {{- end }}
-  
+
   {{- if eq .Values.storage "azure" }}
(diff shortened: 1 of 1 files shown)


canterberry:
Thanks @jthurner! Sorry for sitting on this for so long.

Practically, I can see how the changeset here would reduce some operational friction when not explicitly setting `secrets.haSharedSecret` and attempting to upgrade a previously installed release of the chart.

Philosophically, it doesn't feel right to me for the target cluster state to impact the chart's behavior. It introduces a split source of truth for the value of a secret and makes it more difficult to comprehensively test the chart without applying it against a live cluster.

Given:
 * the core issue is that the default randomly generated value for the HA shared secret (rightfully) forces pods to disruptively cycle, and
 * there is already a solution for this (explicitly set `secrets.haSharedSecret`), and
 * the proposed solution has testability (requires live cluster) and architecture (creates split source of truth) implications...

I'll respectfully abstain from offering my approval on this PR, but I welcome further discussion and review! I'll close this PR after 2 weeks if there is no activity and/or there are no objections.

jthurner:
Thank you for the thorough reply - I can see your point and am closing the PR.

Some more background on motivation: I had bundled docker-registry with [cesanta/docker_auth](https://github.com/cesanta/docker_auth) and keep values files in git. That makes explicitly setting `secrets.haSharedSecret` cumbersome (requiring helm-secrets) and disruption of the registry more frequent (every config change to `docker_auth` will roll the registry deployment if haSharedsecret is auto-generated). Adding a lookup felt less intrusive than bloating the config with another option for referencing an existing haSharedsecret.

But I agree it introduces some obscurity. The umbrella chart approach is less convenient as it looked at first, and the needless pod disruption not that big of a problem in practice. So I am fine with closing this PR.

PR #57, opened 2022-03-04
feat: add logic to support custom liveness and readiness probes

### LINKED ISSUE #54: Add support for custom livenessProbe and readinessProbes
In my case, the default failure count is sometimes too low. I would have wanted an ability to choose probes of my own, as supported by many official helm charts ([filebeat](https://github.com/elastic/helm-charts/blob/main/filebeat/templates/deployment.yaml#L116), for example.)

AbrohamLincoln:
Closes #54

Retains current liveness and readiness probe logic while adding the ability to override the defaults.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -22,3 +22,41 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Create a livenessProbe.
+Allow the default value to be completely overriden by an optional value.
+Retain he original livenessProbe logic.
+*/}}
+{{- define "livenessProbe" -}}
+livenessProbe:
+{{- if .Values.livenessProbe }}
+{{ .Values.livenessProbe | toYaml | indent 2 }}
+{{- else }}
+  httpGet:
+{{- if .Values.tlsSecretName }}
+    scheme: HTTPS
+{{- end }}
+    path: /
+    port: 5000
+{{- end -}}
+{{- end -}}
+
+{{/*
+Create a readinessProbe.
+Allow the default value to be completely overriden by an optional value.
+Retain he original readinessProbe logic.
+*/}}
+{{- define "readinessProbe" -}}
+readinessProbe:
+{{- if .Values.readinessProbe }}
+{{ .Values.readinessProbe | toYaml | indent 2 }}
+{{- else }}
+  httpGet:
+{{- if .Values.tlsSecretName }}
+    scheme: HTTPS
+{{- end }}
+    path: /
+    port: 5000
+{{- end -}}
+{{- end -}}
--- a/values.yaml
+++ b/values.yaml
@@ -168,8 +168,8 @@ autoscaling:
   minReplicas: 1
   maxReplicas: 2
   targetCPUUtilizationPercentage: 60
-  targetMemoryUtilizationPercentage: 60 # available only on Kubernetes ≥1.23 [required "autoscaling/v2"]
-  behavior: {} # available only on Kubernetes ≥1.23 [required "autoscaling/v2"]
+  targetMemoryUtilizationPercentage: 60  # available only on Kubernetes ≥1.23 [required "autoscaling/v2"]
+  behavior: {}  # available only on Kubernetes ≥1.23 [required "autoscaling/v2"]
 #   scaleDown:
 #     stabilizationWindowSeconds: 300
 #     policies:
@@ -213,3 +213,13 @@ extraEnvVars: []
 ## Additional ENV variables to set
 # - name: REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY
 #   value: "/var/lib/example"
+
+# # Custom LivenessProbe config to override default
+# livenessProbe:
+#   tcpSocket:
+#     port: 5000
+
+# # Custom readinessProbe config to override default
+# readinessProbe:
+#   tcpSocket:
+#     port: 5000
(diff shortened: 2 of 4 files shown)


vladciobancai:
Are there any plans for this PR to be merged ?

templates/_helpers.tpl
```diff
@@ -22,3 +22,41 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Create a livenessProbe.
+Allow the default value to be completely overriden by an optional value.
+Retain he original livenessProbe logic.
+*/}}
+{{- define "livenessProbe" -}}
```
canterberry, line 31: For consistency with other chart-specific definitions, include a `docker-registry.` prefix here and on the `readinessProbe` definition.

values.yaml
```diff
@@ -168,8 +168,8 @@ autoscaling:
   minReplicas: 1
   maxReplicas: 2
   targetCPUUtilizationPercentage: 60
-  targetMemoryUtilizationPercentage: 60 # available only on Kubernetes ≥1.23 [required "autoscaling/v2"]
-  behavior: {} # available only on Kubernetes ≥1.23 [required "autoscaling/v2"]
+  targetMemoryUtilizationPercentage: 60  # available only on Kubernetes ≥1.23 [required "autoscaling/v2"]
+  behavior: {}  # available only on Kubernetes ≥1.23 [required "autoscaling/v2"]
```
canterberry, line 172: Spurious whitespace changes here. Please remove.

canterberry (changes requested):
Thanks for contributing! Just made a review pass and have quite minor blocking bits of feedback (just about adding a prefix to the new variable definitions and omitting extraneous whitespace changes). With that, I have no concerns with :+1: and having this merged.

templates/_helpers.tpl
AbrohamLincoln: Added prefix.

values.yaml
AbrohamLincoln: Removes spaces.  Yamllint has a [rule](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.comments:~:text=min%2Dspaces%2Dfrom%2Dcontent%20is%20used%20to%20visually%20separate%20inline%20comments%20from%20content.%20It%20defines%20the%20minimal%20required%20number%20of%20spaces%20between%20a%20comment%20and%20its%20preceding%20content.) that requires two spaces between content and comments by default.

canterberry:
During local manual verification after cloning this branch locally, I expected to be able to install the chart with `--set readinessProbe.httpGet.path=/metrics` to patch just the path of the readiness probe without impacting the rest of its configuration. Instead, I got an error for `missing required field "port"`. As an end-user, I don't know what value I need to provide for that, because port `5000` is hard-coded into the chart as the port on which the container is listening. Further, in the default scenario, the chart decides when to set the `scheme` based on the presence of a TLS secret.

Upon closer inspection, then, and for the above reasons, I'm not sure that a full-on livenessProbe and readinessProbe override is an appropriate solution. It simply requires too much intermingled knowledge of anxilliary parts of the chart.

Based on the ask in #54, a more targeted solution may be more appropriate and less cumbersome. 

To help in narrowing down on a solution:

 * If it the livenessProbe or the readinessProbe (or both) where the failure count is too low? Maybe only one or the other needs to be adjusted. If they're essentially the same (which in the chart currently, they're identical), then it may be a signal for the chart to make them configurable as a set.
 * At what point are the failures happening? If it's as your registry is being launched, then the default `initialDelaySeconds` may be too low for you (and likely others).

Here are the timing and threshold related properties on liveness and readiness probes, for reference:

 * [ ] `failureThreshold`
 * [ ] `initialDelaySeconds`
 * [ ] `periodSeconds`
 * [ ] `successThreshold`
 * [ ] `terminationGracePeriodSeconds`
 * [ ] `timeoutSeconds`

My soft suggestion here would be to, instead of allowing the whole `livenessProbe` and `readinessProbe` blocks to be overridden, to support overriding each of these individual properties above, for each.

A simpler implementation might be to merge any configured livenessProbe into the default one, so the `livenessProbe` *value* becomes a patch rather than a replacement. It's much more powerful than it needs to be, though, to remedy the false-negatives reported in #54, and opens up a higher probability of misconfigured or dysfunctional probes.

values.yaml
canterberry: Ahh, interesting. I've never seen that rule before. Thanks for the context! I didn't know yamllint existed, and I'm glad it does. CI for this repo is currently doing `helm lint --strict`, which doesn't enforce nearly as much as one might naively expect it to. We should add yamllint to this repo's CI and get a PR together to fix any current violations so we have a basis for code style consistency.

canterberry:
@AbrohamLincoln Hey, sorry for what may have come across as shutting down your good work here. I'd really like to address the core issue! Any thoughts or suggestions on how to move forward?

AbrohamLincoln:
Sorry, I've been offline since making that last commit. I made these
changes to work around an edge case I encountered involving PKI
authentication. I ended up completely overriding the probes in my
scenario.  The default settings in the chart seem to be fine for most use
cases. We can add some additional settings if you'd like, but I always
appreciate having a mechanism to allow for advanced configurations.

templates/_helpers.tpl
```diff
@@ -22,3 +22,41 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
+{{ .Values.readinessProbe | toYaml | indent 2 }}
+{{- else }}
+  httpGet:
+{{- if .Values.tlsSecretName }}
+    scheme: HTTPS
+{{- end }}
+    path: /
+    port: 5000
+{{- end -}}
+{{- end -}}
```
edwargix, line 62: nit: add \n

PR #61, opened 2022-06-27
Fix reference in README

joneteus:
https://github.com/twuni/docker-registry.helm/blob/f1583238750b54247c32b3d7cc1e6b72c0dbdc8c/templates/secret.yaml#L17

According to `secret.yaml` `haSharedSecret` should be inside `secrets` object


canterberry (approved):
:+1: Thanks, @joneteus ! Spot-on.

PR #28, opened 2021-09-18
Garbage collector cronjob

kuzaxak:
Please check additional information in commit messages.

Garbage collector job was added to registry [here][1]

[1]: https://github.com/distribution/distribution/pull/2302

PR #62, opened 2022-07-05
:recycle: Deduplicate definitions from deployment and cronjob

### LINKED ISSUE #63: Support garbage collection cronjob
Hi,
Isn't it going to be good idea to have cronjob env where we set cronjob string on which gc command is going to be triggered?

ddelange:
Based on #28 

Closes #63 

Added one commit to bring the branch up to date with latest upstream

ref https://github.com/ddelange/docker-registry.helm/pull/1 for testing purposes, this can be installed with:
```
helm install --namespace ... https://github.com/ddelange/docker-registry.helm/releases/latest/download/helm-chart-docker-registry-2.2.0.tgz
```


ddelange:
cc @canterberry @kuzaxak

1003n40 (dismissed):

ddelange:
@canterberry what do you think? its opt-in, non-breaking, should be safe to merge?

canterberry:
First: thank you! I know you don't have to contribute anything at all, but you did! I and the community are grateful, and I don't show that gratitude early and often enough. So thank you.

Now, let's talk about how we can get this done!

Your general approach seems fine to me -- a new configuration block, clearly documented in the README, and disabled by default. When enabled, there's a cronjob that runs at some configurable frequency, using the registry's own CLI and configuration to perform the garbage collection. Great!

Can you distill that down into a [near-]completely additive changeset?

Rationale: You've done some refactoring here that goes above and beyond, which is great holistically, but in terms of code review, makes it much more difficult to evaluate the impact the changeset will have on existing code. For this chart in particular, I'm very hesitant to review any changes that aren't a slam-dunk, because I'm super cautious about breaking it, and since I'm not a user, personally, I rely on tidy changesets and easy code reviews to build that confidence to ship.

Another small thing: no need to increment the version in Chart.yaml. That's something I do as part of the release process, in case there are multiple PRs ready for merge around the same time.

Let's go! :tada:

ddelange:
Hi @canterberry 👋 

> For this chart in particular, I'm very hesitant to review any changes that aren't a slam-dunk, because I'm super cautious about breaking it, and since I'm not a user, personally, I rely on tidy changesets and easy code reviews to build that confidence to ship.

I can totally understand.

> Can you distill that down into a [near-]completely additive changeset?

If you do a ctrl+F for `include` in the diff of this PR, there will be hits in both `deployment.yaml` and `cronjob.yaml` which require the same blocks. Turning this PR into an additive changeset would basically mean copy-pasting all of `deployment.yaml` into `cronjob.yaml`, and any future changes would need to happen in both places.

If you do a ctrl+F for `define` in the diff of this PR, you'll see the three blocks that were moved from `deployment.yaml` to `_helpers.tpl` (1:1 without changes) so that they can be used by both yamls via `include`.

I don't want to try to convince you to merge a PR you don't feel safe merging, but maybe there's a way of getting you a diff view that helps you review the PR. If you'd open these side by side, and do a visual inspect, maybe the triviality of the change becomes evident and we can get closer to that slam-dunk level? :)

- envs
  - [before](https://github.com/twuni/docker-registry.helm/blob/79c14893afc81c92f72a6f81245aee58e826b0d6/templates/deployment.yaml#L87-L195)
  - [after](https://github.com/twuni/docker-registry.helm/blob/29a0863ee2190d2b214b7625d4c31e9b5618b7e3/templates/_helpers.tpl#L26-L149)
- volumeMounts
  - [before](https://github.com/twuni/docker-registry.helm/blob/79c14893afc81c92f72a6f81245aee58e826b0d6/templates/deployment.yaml#L196-L215)
  - [after](https://github.com/twuni/docker-registry.helm/blob/29a0863ee2190d2b214b7625d4c31e9b5618b7e3/templates/_helpers.tpl#L151-L176)
- volumes
  - [before](https://github.com/twuni/docker-registry.helm/blob/79c14893afc81c92f72a6f81245aee58e826b0d6/templates/deployment.yaml#L228-L256)
  - [after](https://github.com/twuni/docker-registry.helm/blob/29a0863ee2190d2b214b7625d4c31e9b5618b7e3/templates/_helpers.tpl#L178-L211)

I'll revert the version bump in any case. Please let me know what you think!

ddelange:
Would CI help boost confidence? e.g. https://github.com/helm/chart-testing-action

I think their example workflow would shows there's no changes to the rendered deployment.yaml

canterberry:
Thank for your diligence. I can tell you're trying to make it easier for me, and I appreciate the effort! It's not what I need, though, so I'd like to redirect that energy toward achieving our mutual goal of getting this welcome feature delivered.
 
In order to get this PR merged, please make it [near-]completely additive. No refactoring, no formatting changes. Just what is strictly necessary to deliver the feature as described. Only then will I feel comfortable reviewing the changeset. At that point, you might expect some minor feedback or changes requested, but that cycle time would be much shorter and I think we could get this merged quite quickly.

> Just so you know, I **have** done some due diligence here and am not being blindly dismissive! I've tried reviewing the changeset multiple times, even again today, and it's exceedingly difficult to do with any degree of confidence because the changeset is not as identical as you describe. There are many minor deviations -- collectively a non-trivial refactoring. Further, without a comprehensive test suite to evaluate the correctness of all classes of variance in the chart output, an equality check on a single known output is insufficient to provide a satisfactory degree of confidence (although any such test case would be an improvement over nothing at all, and I'd love to have such a suite in this highly configurable chart!)

ddelange:
Hi @canterberry,

I've opened #68 with additional commit fd7fdca.

Once that merges, this PR will become a refactor-PR, with as diff the revert commit d78c010.

Is that what you are aiming for?

canterberry:
@ddelange Once you rebase and/or merge the latest into this PR, I'll take a thorough look. I think it makes sense, conceptually, that the CronJob's pod(s) and the Deployment's pod(s) run with matching environments and volumes. The only potential issue I see coming up in review is whether (and to what extent) any functional changes or regressions might have slipped in. None have yet caught my eye, though.

values.yaml
```diff
@@ -76,7 +76,6 @@ storage: filesystem
 secrets:
   haSharedSecret: ""
   htpasswd: ""
-
```
ddelange:
```suggestion
  htpasswd: ""
```

templates/deployment.yaml
```diff
@@ -14,18 +14,17 @@ spec:
...
   minReadySeconds: 5
   template:
     metadata:
       labels:
         app: {{ template "docker-registry.name" . }}
         release: {{ .Release.Name }}
-        {{- if .Values.podLabels }}
-{{ toYaml .Values.podLabels | indent 8 }}
+        {{- with .Values.podLabels }}
+        {{ toYaml . | nindent 8 }}
```
canterberry, line 27: Can you elaborate on the difference between the before/after here? I'm mainly just curious, since the difference, if any, seems either stylistic or quite nuanced, and I'd like to understand:

```yaml
{{- if .Values.podLabels }}
{{ toYaml .Values.podLabels | indent 8 }}
{{- end }}
```

...vs...

```yaml
{{- with .Values.podLabels }}
{{ toYaml . | nindent 8 }}
{{- end }}
```

Is one form or the other more idiomatic? Is there a subtle whitespace difference in the two outputs? This could affect the way I personally write conditionals like this, hence the endeavor to learn.

README.md
```diff
@@ -117,7 +117,7 @@ their default values.
 | `initContainers`            | Init containers to be created in the pod                                                   | `[]`            |
 | `garbageCollect.enabled`    | If true, will deploy garbage-collector cronjob                                             | `false`         |
 | `garbageCollect.deleteUntagged` | If true, garbage-collector will delete manifests that are not currently referenced via tag | `true` |    |
-| `garbageCollect.schedule`   | CronTab schedule, please use standard crontab format                                        | `0 1 * * *` |  |
+| `garbageCollect.schedule`   | CronTab schedule, please use standard crontab format                                        | `0 1 * * *` |  |
```
ddelange, line 120:
```suggestion
| `garbageCollect.schedule`   | CronTab schedule, please use standard crontab format                                        | `0 1 * * *` |  |
```

templates/deployment.yaml
ddelange: Maybe @kuzaxak can add to the reasoning behind the change, but here are my two cents:

As from [docs](https://helm.sh/docs/chart_template_guide/control_structures/#modifying-scope-using-with) I think `with` will mainly result in shorter statements when accessing multiple objects in a nested structure, and arguably increases readability.

I find the `if` variation in this simple form just as adequate and I don't think performance (multiple lookups) is of any relevance here. Often you won't need the `else` option and the both are interchangeable, where `if` probably feels more familiar for devs coming from functional languages.

canterberry: Thanks for the $0.02! Both versions seem fine. In my view, there doesn't seem to be an adequately convincing argument to use one over the other, but it's also trivial enough and part of a larger changeset that the choice doesn't carry much weight either way. :+1:

canterberry:
To inspect the changes in this PR, I ran the following command from a local clone of this PR's branch:

```bash
$ helm template --namespace meta-namespace --set namespace=target-namespace  --set garbageCollect.enabled=true drc .
```

In both the resulting `CronJob` and `Deployment` manifests, there were a couple of extra newlines preceding the `env` block:

```yaml
              env: 
                
                
                - name: REGISTRY_HTTP_SECRET
                  valueFrom:
                    secretKeyRef:
                      name: drc-docker-registry-secret
                      key: haSharedSecret
                - name: REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY
                  value: "/var/lib/registry"
```

While this doesn't seem to have any functional impact (`helm upgrade --install --dry-run ...` with the same arguments comes back successful), it's something that looks visually off, and I'd like to see that fixed. Other than that, I've finished a review pass and LGTM. I'd like to let it "bake" for a day or two to give others a change to review as well, just in case we've missed something.

templates/_helpers.tpl
```diff
@@ -22,3 +22,190 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
+      key: proxyPassword
+{{- end }}
+
+{{- if .Values.persistence.deleteEnabled }}
+- name: REGISTRY_STORAGE_DELETE_ENABLED
+  value: "true"
+{{- end }}
+
+{{- with .Values.extraEnvVars }}
+        {{- toYaml . | nindent 12 }}
```
ddelange, line 146: need to test, think indentation of `extraEnvVars` will come out wrong
```suggestion
{{- toYaml . }}
```

templates/cronjob.yaml
```diff
@@ -51,135 +51,8 @@ spec:
...
-                  mountPath: "/etc/docker/registry"
-{{- if .Values.tlsSecretName }}
-                - mountPath: /etc/ssl/docker
-                  name: tls-cert
-                  readOnly: true
-{{- end }}
-{{- with .Values.extraVolumeMounts }}
-                {{- toYaml . | nindent 16 }}
-{{- end }}
+              env: {{ include "docker-registry.envs" . | nindent 16 }}
```
ddelange, line 54:
```suggestion
              env: {{- include "docker-registry.envs" . | nindent 16 }}
```

templates/cronjob.yaml
```diff
@@ -51,135 +51,8 @@ spec:
...
+              volumeMounts: {{ include "docker-registry.volumeMounts" . | nindent 16 }}
```
ddelange, line 55:
```suggestion
              volumeMounts: {{- include "docker-registry.volumeMounts" . | nindent 16 }}
```

templates/cronjob.yaml
```diff
@@ -190,33 +63,5 @@ spec:
...
-                name: {{ template "docker-registry.fullname" . }}-config
-{{- if .Values.tlsSecretName }}
-            - name: tls-cert
-              secret:
-                secretName: {{ .Values.tlsSecretName }}
-{{- end }}
-{{- with .Values.extraVolumes }}
-            {{- toYaml . | nindent 12 }}
-{{- end }}
+          volumes: {{ include "docker-registry.volumes" . | nindent 12 }}
```
ddelange, line 66:
```suggestion
          volumes: {{- include "docker-registry.volumes" . | nindent 12 }}
```

templates/deployment.yaml
```diff
@@ -70,187 +68,28 @@ spec:
...
-{{- if .Values.tlsSecretName }}
-        - name: tls-cert
-          secret:
-            secretName: {{ .Values.tlsSecretName }}
-{{- end }}
-{{- with .Values.extraVolumes }}
-        {{- toYaml . | nindent 8 }}
-{{- end }}
+          resources: {{ toYaml .Values.resources | nindent 12 }}
+          env: {{ include "docker-registry.envs" . | nindent 12 }}
```
ddelange, line 84:
```suggestion
          env: {{- include "docker-registry.envs" . | nindent 12 }}
```

templates/deployment.yaml
```diff
@@ -70,187 +68,28 @@ spec:
...
+          volumeMounts: {{ include "docker-registry.volumeMounts" . | nindent 12 }}
```
ddelange, line 85:
```suggestion
          volumeMounts: {{- include "docker-registry.volumeMounts" . | nindent 12 }}
```

templates/deployment.yaml
```diff
@@ -70,187 +68,28 @@ spec:
...
+      {{- if .Values.nodeSelector }}
+      nodeSelector: {{ toYaml .Values.nodeSelector | nindent 8 }}
+      {{- end }}
+      {{- if .Values.affinity }}
+      affinity: {{ toYaml .Values.affinity | nindent 8 }}
+      {{- end }}
+      {{- if .Values.tolerations }}
+      tolerations: {{ toYaml .Values.tolerations | nindent 8 }}
+      {{- end }}
+      volumes: {{ include "docker-registry.volumes" . | nindent 8 }}
```
ddelange, line 95:
```suggestion
      volumes: {{- include "docker-registry.volumes" . | nindent 8 }}
```

ddelange (commented):
I think this should do the trick [ref](https://v2.helm.sh/docs/charts_tips_and_tricks/)

templates/deployment.yaml
canterberry: Just tried this locally (for this and the other `{{- ... }}` suggested changes), and got the same result.

ddelange: ah I'm sorry to waste your time, I'm on mobile. will push a fix from my machine tomorrow 🤙

canterberry: I figured you were just taking notes. No worries! Tag me when you're ready for :eyes: .

templates/_helpers.tpl
ddelange: confirmed with
```
helm template --debug . --set extraEnvVars[0].name=TEST_NAME --set extraEnvVars[1].value=TEST_VALUE
```

templates/deployment.yaml
kuzaxak: Hi,

As you said with works in the same way as in Python. If you have nested object somewhere and you want to change pointer for the current context `.` you can use `with`.

But keep that in mind if you need to access anything outside of the context you need to use root context via `$.`.

Main goal is reduce duplicating nested path, as an example:

To render something like `.worker.ingresses.first.annotations` it would make sense to use `with`

templates/_helpers.tpl
```diff
@@ -22,3 +22,190 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
+    secretKeyRef:
+      name: {{ template "docker-registry.fullname" . }}-secret
+      key: haSharedSecret
+
+{{- if .Values.tlsSecretName }}
+- name: REGISTRY_HTTP_TLS_CERTIFICATE
+  value: /etc/ssl/docker/tls.crt
+- name: REGISTRY_HTTP_TLS_KEY
+  value: /etc/ssl/docker/tls.key
+{{- end }}
```
kuzaxak, line 47: To remove next newline (if it is important, but won't change anything) you can use something like that:

`{{- end -}}`

templates/_helpers.tpl
```diff
@@ -22,3 +22,190 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
+- name: REGISTRY_STORAGE_S3_ENCRYPT
+  value: {{ .Values.s3.encrypt | quote }}
+{{- end }}
+
+{{- if .Values.s3.secure }}
+- name: REGISTRY_STORAGE_S3_SECURE
+  value: {{ .Values.s3.secure | quote }}
+{{- end }}
+
+{{- else if eq .Values.storage "swift" }}
```
kuzaxak, line 108: If you have time you can split this chunk of variables into separate define sections, one per type, and then render them inside that define.

In that case you will achieve function like style where every Backend storage have isolated function to render their env variables.

Just a note, existing solution is fine as well.

templates/_helpers.tpl
kuzaxak, line 146: You don't need indention before line with `nindent` because nindent insert new line and the indentent content

kuzaxak (commented):
Small notes from me. Generally LGTM. Thank you for picking up my PR :)

ddelange:
env whitespaces fixed in 52a5d4b

tested with
```
helm template --debug . \
  --set serviceAccount.create=true \
  --set priorityClassName=high \
  --set extraEnvVars[0].name=TEST_NAME \
  --set extraEnvVars[0].value=TEST_VALUE \
  --set secrets.htpasswd=abc \
  --set tlsSecretName=abc \
  --set garbageCollect.enabled=true \
  --set proxy.enabled=true \
  --set storage=s3 /
  --set secrets.s3.secretKey=abc /
  --set secrets.s3.accessKey=def /
  --set s3.region=us-42 /
  --set s3.bucket=abc /
  --set s3.encrypt=abc
```

templates/_helpers.tpl
ddelange: 52a5d4b

templates/_helpers.tpl
ddelange: 52a5d4b

ddelange:
I think we're not there yet. This might be a sensible Github Workflow to comment (and edit that comment) with a diff on a PR:

```sh
alias helm_test_template='helm template --debug \
  --set serviceAccount.create=true \
  --set priorityClassName=high \
  --set podAnnotations.test=annotation \
  --set extraEnvVars[0].name=TEST_NAME \
  --set extraEnvVars[0].value=TEST_VALUE \
  --set secrets.htpasswd=abc \
  --set tlsSecretName=abc \
  --set garbageCollect.enabled=true \
  --set proxy.enabled=true \
  --set storage=s3 /
  --set secrets.s3.secretKey=abc /
  --set secrets.s3.accessKey=def /
  --set s3.region=us-42 /
  --set s3.bucket=abc /
  --set s3.encrypt=abc'
diff -u <(helm_test_template https://github.com/twuni/docker-registry.helm/archive/refs/heads/main.tar.gz) <(helm_test_template .)
```
gives in the ` ```diff ` syntax:
```diff
--- /dev/fd/63	2022-08-10 10:48:29.000000000 +0200
+++ /dev/fd/62	2022-08-10 10:48:29.000000000 +0200
@@ -12,7 +12,7 @@
     release: RELEASE-NAME
 type: Opaque
 data:
-  haSharedSecret: "S2pTbHFaeEtCZ2ppTHY3bQ=="
+  haSharedSecret: "TnZBakJNSWVpbXRKeWFSUw=="
   proxyUsername: ""
   proxyPassword: ""
 ---
@@ -99,7 +99,7 @@
         release: RELEASE-NAME
       annotations:
         checksum/config: 581c8cf6d59449d6c89ebf8ec1bef05751a9c8e281ca31f99c4e0291e054bfaf
-        checksum/secret: 153f24ee9a430e533197ba266c83c185b4f95d38cd6ee547c93eca76a1969b85
+        checksum/secret: 25156342abc38ff3b671e698c348aa9d5726d62a1a9bcb539437febc7e5b8ccc
     spec:
       securityContext:
         fsGroup: 1000
@@ -122,9 +122,9 @@
             httpGet:
               path: /
               port: 5000
-          resources:
+          resources:
             {}
-          env:
+          env:
             - name: REGISTRY_HTTP_SECRET
               valueFrom:
                 secretKeyRef:
@@ -132,14 +132,14 @@
                   key: haSharedSecret
             - name: REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY
               value: "/var/lib/registry"
-          volumeMounts:
-            - name: data
-              mountPath: /var/lib/registry/
+          volumeMounts:
             - name: "RELEASE-NAME-docker-registry-config"
               mountPath: "/etc/docker/registry"
-      volumes:
-        - name: data
-          emptyDir: {}
+            - name: data
+              mountPath: /var/lib/registry/
+      volumes:
         - name: RELEASE-NAME-docker-registry-config
           configMap:
             name: RELEASE-NAME-docker-registry-config
+        - name: data
+          emptyDir: {}
```

ddelange:
^ opened #70

canterberry (approved):
👍 LGTM. Whitespace looks good now, with my limited-scope smoke test.

canterberry:
Released in [v2.2.1](https://github.com/twuni/docker-registry.helm/releases/tag/v2.2.1).

PR #68, opened 2022-08-05
Add option to enable garbage collector

ddelange:
Additive version of #62


templates/cronjob.yaml
```diff
@@ -0,0 +1,202 @@
+{{- if .Values.garbageCollect.enabled }}
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: {{ template "docker-registry.fullname" . }}-garbage-collector
+  labels:
+    app: {{ template "docker-registry.name" . }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
```
canterberry, line 10: Missing `metadata.namespace` (see https://github.com/twuni/docker-registry.helm/blob/main/templates/configmap.yaml#L5 for reference)

templates/cronjob.yaml
```diff
@@ -0,0 +1,202 @@
+{{- if .Values.garbageCollect.enabled }}
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: {{ template "docker-registry.fullname" . }}-garbage-collector
+  labels:
+    app: {{ template "docker-registry.name" . }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
```
canterberry, line 8: :bulb: **Note:** Looks like `metadata.labels.chart` isn't being set consistently across all resources in the chart. The labels lack prefixes, anyway, making them problematic for some observability tooling (e.g: Elasticsearch will drop log events that include a field with a type conflicting with the index's field type mapping, and the `app` label conflicts with every one of Kubernetes's [recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)).

No changes requested, just making a note of this for reference.

values.yaml
```diff
@@ -76,6 +77,9 @@ storage: filesystem
 secrets:
   haSharedSecret: ""
   htpasswd: ""
+  azure: {}
+  s3: {}
+  swift: {}
```
canterberry, line 82: The changes to these three lines seem spurious to me. Neither `azure` nor `swift` is referenced in this changeset, and `s3` is only referenced when `storage: s3`. While defaulting these to objects may remedy some other issue, in this PR they are spurious.

values.yaml
```diff
@@ -101,6 +106,7 @@ secrets:
 #  encrypt: false
 #  secure: true
 
+swift: {}
```
canterberry, line 109: Spurious change here. Please remove (or address in a separate PR).

values.yaml
```diff
@@ -92,6 +96,7 @@ secrets:
 #     username: ""
 #     password: ""
 
+s3: {}
```
canterberry, line 99: Spurious change here. Please remove or address in a separate PR.

templates/cronjob.yaml
```diff
@@ -0,0 +1,202 @@
...
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  schedule: {{ .Values.garbageCollect.schedule | quote }}
+  jobTemplate:
+    spec:
```
canterberry, line 14: Missing `spec.jobTemplate.metadata` block -- jobs created will be missing labels.

In all fairness, the labels on resources across this chart aren't consistent with one another anyway and could use a round of cleanup.

Soft suggestion to add metadata here with consistent labels, but because of the above, not a blocking issue for me.

templates/cronjob.yaml
```diff
@@ -0,0 +1,202 @@
...
+      template:
+        spec:
```
canterberry, line 16: Missing `spec.jobTemplate.spec.template.spec.serviceAccountName` (see https://github.com/twuni/docker-registry.helm/blob/main/templates/deployment.yaml#L37 for reference).

canterberry: Also missing `spec.jobTemplate.spec.template.spec.priorityClassName` (see https://github.com/twuni/docker-registry.helm/blob/main/templates/deployment.yaml#L44 for reference), although I'm not sure whether the intention carries over for this job as well as the registry pod(s).

Soft suggestion to add that in here, but if there's a good reason to omit it, non-blocking. I don't understand how `priorityClassName` is in use today, so can't make an informed judgment on that myself.

canterberry (changes requested):
Thanks for pushing this through, @ddelange !

Overall looks great! All of my feedback is quite minor. I've noted anything non-blocking where relevant.

I invite anyone else to review as well, in case there's something we missed.

I cloned locally and ran `helm template ...` with a basic configuration as a smoke test (with and without `--set garbageCollect.enabled=true`). :ok_hand:

templates/cronjob.yaml
ddelange: copying https://github.com/ddelange/docker-registry.helm/blob/d78c010ca556a6bfc7650b575180ddf4d0fadc44/templates/deployment.yaml#L22-L34

templates/cronjob.yaml
ddelange: 👍 for now, I'll leave label/annotation consistency out of this PR

ddelange:
@canterberry thanks for the review, I've pushed the changes also to #62

templates/cronjob.yaml
```diff
@@ -0,0 +1,222 @@
...
+              - name: REGISTRY_PROXY_PASSWORD
+                valueFrom:
+                  secretKeyRef:
+                    name: {{ if .Values.proxy.secretRef }}{{ .Values.proxy.secretRef }}{{ else }}{{ template "docker-registry.fullname" . }}-secret{{ end }}
+                    key: proxyPassword
+{{- end }}
+{{- if .Values.persistence.deleteEnabled }}
+              - name: REGISTRY_STORAGE_DELETE_ENABLED
+                value: "true"
+{{- end }}
```
canterberry, line 159: 💡 **Note:** This environment variable (and corresponding value in values.yaml) is introduced in this PR, so in subsequent refactoring, just need to be careful of regression.

values.yaml
```diff
@@ -67,6 +67,7 @@ persistence:
   enabled: false
   size: 10Gi
   # storageClass: '-'
+  deleteEnabled: false
```
canterberry, line 70: Missing documentation for `persistence.deleteEnabled` in **README.md**. At first look, this seems potentially spurious, but I can see how it may be necessary for garbage collection to function.

ddelange: potentially spurious indeed. maybe @kuzaxak added it simply because it was [missing](https://github.com/twuni/docker-registry.helm/search?q=deleteenabled) from values.yaml? 

potentially similar to the other spurious comments above

ddelange:
```suggestion
```

canterberry (approved):
Thanks for the link, @ddelange , showing where `persistence.deleteEnabled` is referenced. LGTM. I'll merge as soon as GitHub's checks are green.

canterberry:
Thanks for the persistence, @ddelange , and @kuzaxak for getting this started! This has now been shipped in v2.2.0.

ddelange:
Awesome, thanks to you too for taking your time for reviewing and writing your comments the way you do. Very much appreciated, taking it with in my napsack!

Raboo:
Will this garbage collector delete docker buildx cache?

canterberry:
> Will this garbage collector delete docker buildx cache?

AFAIK, `docker build` and `docker buildx` caches are only on the local machine where the image(s) are built or pulled. The registry itself wouldn't have a cache, per se. There's more information on garbage collection in the [official docs](https://docs.docker.com/registry/garbage-collection/) -- to my understanding, it's just deleting blobs that aren't being referenced anymore.

Raboo:
@canterberry buildx caches to an registry using --cache-to (https://docs.docker.com/engine/reference/commandline/buildx_build/#cache-to). This is especially useful when you use automated pipelines where you don't have any local cache.

But I guess that the only way for me to find out is to try it, since it doesn't say anything about the cache in the garbage collection documentation.

ddelange:
This feature can be perfectly combined with `type=registry` caching. It was our initial reason for pushing the PR :) Let me explain:

When your builder pushes cache to the `ref=bla` tag for this image (e.g. `type=registry,ref=buildcache`), it overwrites the previous `buildcache` image tag that was already there. That previous image becomes untagged (can now only be fetched/referenced with `@sha` notation), and thats exactly the type of images the garbage collector will delete from the storage backend using the `deleteUntagged` option.

So with this feature enabled, your registry won't explode in size anymore when your workflows like to regularly overwrite existing tags.

PR #70, opened 2022-08-10
:construction_worker: Add sticky helm diff PR comment

ddelange:
ref https://github.com/twuni/docker-registry.helm/pull/62#issuecomment-1210366076

example https://github.com/ddelange/docker-registry.helm/pull/5#issuecomment-1210654878

### DIFF
--- a/.github/workflows/pr_diff.yaml
+++ b/.github/workflows/pr_diff.yaml
@@ -11,6 +11,7 @@ jobs:
       - name: Run diff
         id: diff
         run: |
+          shopt -s expand_aliases
           alias helm_test_template='helm template --debug \
             --set serviceAccount.create=true \
             --set priorityClassName=high \
--- a/.github/workflows/pr_diff.yaml
+++ b/.github/workflows/pr_diff.yaml
@@ -11,26 +11,26 @@ jobs:
       - name: Run diff
         id: diff
         run: |
-          shopt -s expand_aliases
-          alias helm_test_template='helm template --debug \
-            --set serviceAccount.create=true \
-            --set priorityClassName=high \
-            --set podAnnotations.test=annotation \
-            --set extraEnvVars[0].name=TEST_NAME \
-            --set extraEnvVars[0].value=TEST_VALUE \
-            --set secrets.htpasswd=abc \
-            --set tlsSecretName=abc \
-            --set garbageCollect.enabled=true \
-            --set proxy.enabled=true \
-            --set storage=s3 /
-            --set secrets.s3.secretKey=abc /
-            --set secrets.s3.accessKey=def /
-            --set s3.region=us-42 /
-            --set s3.bucket=abc /
-            --set s3.encrypt=abc'
+          OPTIONS=(
+            --set serviceAccount.create=true
+            --set priorityClassName=high
+            --set podAnnotations.test=annotation
+            --set extraEnvVars[0].name=TEST_NAME
+            --set extraEnvVars[0].value=TEST_VALUE
+            --set secrets.htpasswd=abc
+            --set tlsSecretName=abc
+            --set garbageCollect.enabled=true
+            --set proxy.enabled=true
+            --set storage=s3
+            --set secrets.s3.secretKey=abc
+            --set secrets.s3.accessKey=def
+            --set s3.region=us-42
+            --set s3.bucket=abc
+            --set s3.encrypt=abc
+          )
           # https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings
           echo 'HELM_DIFF<<EOF' >> $GITHUB_ENV
-          diff -u <(helm_test_template ${{ github.server_url }}/${{ github.repository }}/archive/refs/heads/${{ github.base_ref }}.tar.gz) <(helm_test_template .) >> $GITHUB_ENV
+          diff -u <(helm template --debug ${OPTIONS[@]} ${{ github.server_url }}/${{ github.repository }}/archive/refs/heads/${{ github.base_ref }}.tar.gz) <(helm template --debug ${OPTIONS[@]} .) >> $GITHUB_ENV
           echo 'EOF' >> $GITHUB_ENV
       - uses: marocchino/sticky-pull-request-comment@v2
         with:
--- a/.github/workflows/pr_diff.yaml
+++ b/.github/workflows/pr_diff.yaml
@@ -30,7 +30,7 @@ jobs:
           )
           # https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings
           echo 'HELM_DIFF<<EOF' >> $GITHUB_ENV
-          diff -u <(helm template --debug ${OPTIONS[@]} ${{ github.server_url }}/${{ github.repository }}/archive/refs/heads/${{ github.base_ref }}.tar.gz) <(helm template --debug ${OPTIONS[@]} .) >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


ddelange:
should start working once this workflow lives in `twuni:main` [ref](https://github.com/marocchino/sticky-pull-request-comment#error-resource-not-accessible-by-integration)

ddelange:
tested here https://github.com/ddelange/docker-registry.helm/pull/2

ddelange:
now with filenames in the diff, and collapsed the markdown https://github.com/ddelange/docker-registry.helm/pull/5

canterberry (commented):
Thanks, this is great! It will save time on PR review, for sure.

I have a couple of concerns, which may or may not require adjustments. Just talking points to discuss before merge:

 1. marocchino/sticky-pull-request-comment is a privately maintained GitHub action, so I'm concerned about introducing a dependency on it. It creates a single point of failure and a heightened security risk compared to, say, a more popular and commercially-backed/community-led action. Not a blocking issue, but just wanted to bring it up.
 2. If the action is failing on this PR, what evidence supports that it would succeed on subsequent PRs? Are any repo configuration changes necessary in order to support the action-initiated PR comment? I can see it passing on your fork, but I don't yet understand why it passes there but fails here. I suspect it's because the branch the PR is based on is local to the target branch (vs any PRs opened against this repo which are coming from forks).

While (1) is not something I'm that concerned about, I'd like to understand (2) better to avoid potentially breaking mergability of PRs in this repo, although the risk is low, because if that were to happen, it would be trivial to fix by reverting this PR.

.github/workflows/pr_diff.yaml
```diff
@@ -0,0 +1,48 @@
+name: PR Diff for Helm chart
+on:
+  pull_request:
```
ddelange, line 3:
```suggestion
  pull_request_target:
```

.github/workflows/pr_diff.yaml
```diff
@@ -0,0 +1,48 @@
...
+            --set s3.bucket=abc
+            --set s3.encrypt=abc
+          )
+          helm template --debug ${OPTIONS[@]} --output-dir before ${{ github.server_url }}/${{ github.repository }}/archive/refs/heads/${{ github.base_ref }}.tar.gz
+          helm template --debug ${OPTIONS[@]} --output-dir after .
+          # https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings
+          echo 'HELM_DIFF<<EOF' >> $GITHUB_ENV
+          echo "$(diff -ur before after)" >> $GITHUB_ENV
+          echo 'EOF' >> $GITHUB_ENV
+      - uses: marocchino/sticky-pull-request-comment@v2
```
ddelange, line 37: v2.2.0 as SHA:
```suggestion
      - uses: marocchino/sticky-pull-request-comment@39c5b5d
```

ddelange (commented):
Very valid concerns!

1) could be remedied by pointing to a commit hash instead of a major version
2) is probably a valid issue ref https://github.com/keptn-contrib/artifacthub/issues/89. I think you might be right that in the current state it won't work for any new PRs coming from forks, even when this code would live on `twuni:main`. I encountered the 'not on master' behaviour in a different setting, and must have gotten them confused. I think it had to do with the `release` trigger, not `pull_request`. Fix would be to run on `pull_request_target` trigger.

.github/workflows/pr_diff.yaml
ddelange: We can bar this off with [permissions](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs):
```suggestion
  pull_request_target:
  
permissions:
  pull-requests: write
```

canterberry:
Thanks for (1) updating to the commit hash, and for the link on (2). [This comment](https://github.com/keptn-contrib/artifacthub/issues/89#issuecomment-1050613565) points out three potential solutions, and unfortunately, neither of them are viable. Granting the Action read/write repository permissions with access to secrets even when running on a fork is a [non-starter](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/):

> TL;DR: Combining pull_request_target workflow trigger with an explicit checkout of an untrusted PR is a dangerous practice that may lead to repository compromise.

The **value** provided by this PR, however, is great enough that I'd still love to see it land in some form. Is there another way the diff can be presented or attached? I'm not familiar with GitHub Actions (my experience is mostly CircleCI and GitLab) -- is there something like Artifacts or Reports that can be attached to the action's result?

.github/workflows/pr_diff.yaml
```diff
@@ -0,0 +1,52 @@
+name: PR Diff for Helm chart
+on:
+  pull_request:
+  pull_request_target:
+  
+permissions:
+  pull-requests: write
```
ddelange, line 7: I think this is the crux. From their [docs](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs), all other permissions should flip to none as per this line. If an attacker would remove this one, in order to gain full permissions from `pull_request_target`, his attack would fail because changes to `on:` on `permissions:` 
blocks should only go into effect already on the PR if the change is not coming from a fork. This `on:` block is already verfied as per this PR (I can't elevate). The first part, so I expect similar behaviour for the `permissions:` block.

More info on the triggers, here in context of a `docker build --push`, where the push should fail in PRs coming from the fork (no access to Github Secrets): https://github.com/pelias/docker/issues/264#issuecomment-1212424894

.github/workflows/pr_diff.yaml
```diff
@@ -0,0 +1,52 @@
...
+          helm template --debug ${OPTIONS[@]} --output-dir before ${{ github.server_url }}/${{ github.repository }}/archive/refs/heads/${{ github.base_ref }}.tar.gz
+          helm template --debug ${OPTIONS[@]} --output-dir after .
+          # https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings
+          echo 'HELM_DIFF<<EOF' >> $GITHUB_ENV
+          echo "$(diff -ur before after)" >> $GITHUB_ENV
+          echo 'EOF' >> $GITHUB_ENV
+      - uses: marocchino/sticky-pull-request-comment@39c5b5dc7717447d0cba270cd115037d32d28443
+        with:
+          message: |
+            Running a `helm template` smoketest on this branch results in the following diff against `${{ github.base_ref }}`:
```
ddelange, line 44:
```suggestion
            Running a `helm template` smoketest on commit ${{ github.ref }} results in the following diff against `${{ github.base_ref }}`:
```

.github/workflows/pr_diff.yaml
canterberry: Alright. There aren't currently any secrets on this repo, since there's no third-party service involved in CI nor automated release/deployment stages. Limiting the permissions to `pull-requests: write` seems reasonable, although still seems potentially broad (i.e: does it mean the action can write to *any* PR against the target?). It's not a significant enough risk, to me, to block from merging, though. Let's try it.

canterberry (approved):
👍 Although the newly introduced action is currently failing, it seems probable that is because it's trying to do something the target hasn't authorized. Post-merge, the target will have authorized the action, so in theory will start working. If not, it's trivial to revert.

canterberry:
Released in [v2.2.1](https://github.com/twuni/docker-registry.helm/releases/tag/v2.2.1).

.github/workflows/pr_diff.yaml
ddelange: > does it mean the action can write to any PR against the target?

I wished GH docs were more elaborate here...

PR #72, opened 2022-08-17
Upgrade default image version from 2.7.1 to 2.8.1

edwargix:
Includes changes from releases 2.8.0 and 2.8.1, which mostly contain bugfixes:

- https://github.com/distribution/distribution/releases/tag/v2.8.0
- https://github.com/distribution/distribution/releases/tag/v2.8.1

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 description: A Helm chart for Docker Registry
 name: docker-registry
 version: 2.2.1
-appVersion: 2.7.1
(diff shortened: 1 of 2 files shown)


canterberry:
cc @ddelange I'm not sure the PR diff action is doing the correct thing here -- the diff doesn't look right, and the refs being compared seem to be the same. I'll look into it.

canterberry (approved):
:+1: LGTM. Thanks, @edwargix!

PR #87, opened 2022-11-22
feat: add statefuleset support on v2.2.2

nicop311:
Adding support for `StatefuleSet` on v2.2.2.

## Source of inspiration

The [grafana helm chart](https://github.com/grafana/helm-charts/tree/main/charts/grafana) supports both StatefuleSet and Deployment.

## Release Notes

Adding support for deploying the container registry using a Kubernetes `StatefuleSet` manifest object with dedicated per-sts PVC. Compatible with podAntiAffinity.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -187,7 +187,7 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
       path: htpasswd
 {{- end }}
 
-{{- if eq .Values.storage "filesystem" }}
+{{- if (and (eq .Values.storage "filesystem") (not .Values.useStatefulSet)) }}
 - name: data
   {{- if .Values.persistence.enabled }}
   persistentVolumeClaim:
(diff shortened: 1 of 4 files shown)


wuestkamp:
This would be awesome! We would like to run multiple replaces with separate volumes

nicop311:
@wuestkamp This is indeed the purpose of my modifications :slightly_smiling_face: ! It is still a draft as I needed to make sure the HELM chart has 2 different behaviors with `Deployment` and `StatefuleSet`. For instance, I added in file [templates/_helpers.tpl](https://github.com/twuni/docker-registry.helm/pull/87/commits/f8ae9d438b0d58dc0a54d16e811cb02fba7f2eba#diff-87d68c754766af8e2e930e653be7e4b75fa0c8bdb187cb1bec293f265d9159ff) I added a line `{{- if (and (eq .Values.storage "filesystem") (not .Values.useStatefulSet)) }}` which prevent creating a dedicated PVC manifest when using `StatefuleSet`.

I did a couple of other modifications like this one in other files, like `pvc.yaml`.

I will soon update the original message with a release note.

nicop311:
I cancel this PR 87 and create a new [PR 89](https://github.com/twuni/docker-registry.helm/pull/89), with the same content but using a corporate process and account.

PR #93, opened 2023-01-26
fix: allow to set emptyDir sizeLimit

pkrzaczkowski-hippo:
This PR allows setting ephemeral storage sizeLimit.

Right now if no PVC storage is defined, an emptyDir volume is created without any size limits. This can cause the storage of the node to run out.

This PR adds an option to limit ephemeral storage the same way as we limit PVC claim storage.

https://kubernetes.io/docs/concepts/storage/volumes/#emptydir

Default value is set to "0" - which is unlimited

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ their default values.
 | `autoscaling.behavior`      | (Kubernetes ≥1.23) Configurable scaling behavior                                           | `{}`            |
 | `priorityClassName      `   | priorityClassName                                                                          | `""`            |
 | `storage`                   | Storage system to use                                                                      | `filesystem`    |
+| `emptydir.size`             | Amount of space to claim for emptyDir ephemeral storage                                    | `0`             |
 | `tlsSecretName`             | Name of secret for TLS certs                                                               | `nil`           |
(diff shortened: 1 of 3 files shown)


pkrzaczkowski-hippo:
@canterberry

canterberry (approved):
One minor quibble, `emptydir` should be `emptyDir`, else it's a typo-in-waiting for someone. Non-blocking, though. Approved.

canterberry:
Thank you!

vyas-n:
Hi @canterberry,

Do you still want signed commits to be a requirement for contributions moving forward?

For example, this PR could be merged without the requirement.

vyas-n:
Alternatively @pkrzaczkowski-hippo,

Would you be able to sign your commits using a GPG key and push it up to this branch?

There's a guide on GitHub on how to sign commits: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

As well as this guide for how to sign and push up previous commits: https://stackoverflow.com/a/15667644/8401320

Once you've resigned your commit, you can push it up with a `git push -f`

Issue #97, opened 2023-03-03
`s3.regionEndpoint` requires scheme


0xDEC0DE:
If configuring an S3-compatible object store with an endpoint that is not AWS, the transport scheme (`http`/`https`) must be provided, or the server will hang on start while logging no errors, and the liveness checks will eventually kill it, and force the pod into a crash/restart loop.

The docs provided in the `values.yaml` do not include a scheme, which leads to confusion.

# Steps to reproduce
In the `config.yaml`:
```
docker-registry:
  storage: s3
  s3:
    region: my-region
    regionEndpoint: s3.compatible.storagedevice.local
    bucket: some-bucket
```
Deploy the chart

# Expected result

The pod comes up and provides a registry

# Actual behavior

The pod cannot actually talk to the S3 service, so it silently hangs until killed by the livenessProbe checks.  Hilarity ensues.

# Workaround

Provide the transport scheme when defining the endpoint:
```
    regionEndpoint: http://s3.compatible.storagedevice.local/
```
The provided `values.yaml` contraindicates this.

vyas-n:
Sorry for the delay in reviewing the issue,

As of the current version of the helm chart (from my personal experience), the transport scheme isn't necessary for that helm value. I'm able to use Ceph's S3 Object Store endpoint without specifying the scheme.

`s3.regionEndpoint: ceph-objectstore.<my-domain>`

This value is passed into the registry container via the environment variable `REGISTRY_STORAGE_S3_REGIONENDPOINT`: https://github.com/twuni/docker-registry.helm/blob/d6bc315183422f0782d7537b2c93fbf09362fce1/templates/_helpers.tpl#L87

While the upstream container doesn't specify either, it appears that they do include the transportation scheme in their example values: https://distribution.github.io/distribution/about/configuration/

Since the ambiguity is confusing, I'll vote in favor of your change to specify the scheme in the endpoint URL. I'll do a quick test to confirm and post my findings in your PR #98.

PR #98, opened 2023-03-03
docs: Document that `s3.regionEndpoint` requires a transport scheme

0xDEC0DE:
The `regionEndpoint` value does not work at all without one, so provide a nice hint that it is required in the vendor-provided materials.

Fixes Issue: #97

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -97,7 +97,7 @@ secrets:
 # Options for s3 storage type:
 # s3:
 #  region: us-east-1
-#  regionEndpoint: s3.us-east-1.amazonaws.com
+#  regionEndpoint: https://s3.us-east-1.amazonaws.com
 #  bucket: my-bucket
 #  rootdirectory: /object/prefix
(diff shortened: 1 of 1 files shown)


0xDEC0DE:
As you can see, I consider this a documentation issue, not an issue with the chart itself.  A simple hint like this would have steered me towards the correct configuration immediately.

vyas-n (approved):
Thanks for your contribution, I confirmed that (at least using Ceph S3 Object Store) the endpoint works with or without the transportation scheme.

But as I stated in the issue #97, adding the scheme will reduce the confusion for users and make our helm chart documentation match the upstream container.

0xDEC0DE:
@vyas-n Apologies, my commit signing setup was a bit of a mess.  That's all rectified now, and you ought be able to re-enable auto-merge for this PR.

vyas-n:
Thanks @0xDEC0DE!

PR #100, opened 2023-03-15
🐞 Fix secrets.s3.accessKey and secretKey

fungiboletus:


fungiboletus:
My bad, I didn't understand that these settings were supposed to contain the actual secrets and not the key reference of an existing secret.

PR #101, opened 2023-03-25
Add sidecars support

fgiannetti:
Add the ability to specify sidecar containers.
The motivation on this change is because I need to implement a whitelist for the registry (in proxy mode) and as the registry image does not support a whitelist configuration, the alternative is running a reverse proxy in the pod to reject the non-whitelisted images
The reverse proxy could run in other pod, however, allowing sidecars to be set will help whoever needs it.
Hope it helps!


0xDEC0DE:
This looks promising.  A terrific use case for this would be running [docker_auth](https://github.com/cesanta/docker_auth.git) as a sidecar, instead of as a separate pod.

Do you think you could update the docs with instructions on getting that working?

PR #102, opened 2023-03-28
Add missing fields for Security context  and secrets

erikfuego:
Due to PodSecurity policies in place after CIS hardening my cluster, I had to add the following to the pod or container security context:
```
runAsNonRoot: true
seccompProfile:
  type: "RuntimeDefault"
allowPrivilegeEscalation: false
capabilities:
  drop:
  - ALL
```
With the current securityContext implementation, only `securityContext.fsGroup` and `securityContext.runAsUser` were included. This change adds the ability to include everything in the `securityContext` provided. It also removes the `enabled` parameter which is unnecessary to me because we can just check if anything in the `securityContext` exists.

I added a `containerSecurityContext` to help differentiate between adding these fields for a container or the service itself. If `securityContext` is added for the service, PodSecurity fails because it specifically asks to be added to the container level.

This also adds a missing field for secrets called `secretRef` for s3.

### DIFF
--- a/templates/deployment.yaml
+++ b/templates/deployment.yaml
@@ -43,9 +43,7 @@ spec:
       priorityClassName: "{{ .Values.priorityClassName }}"
       {{- end }}
       {{- if .Values.securityContext.enabled }}
-      securityContext:
-        fsGroup: {{ .Values.securityContext.fsGroup }}
-        runAsUser: {{ .Values.securityContext.runAsUser }}
+      securityContext: {{ omit .Values.securityContext "enabled" | toYaml | nindent 8 }}
       {{- end }}
       {{- with .Values.initContainers }}
       initContainers:
@@ -82,6 +80,9 @@ spec:
               port: 5000
           resources: {{ toYaml .Values.resources | nindent 12 }}
           env: {{ include "docker-registry.envs" . | nindent 12 }}
+          {{- if .Values.containerSecurityContext.enabled }}
+          securityContext: {{ omit .Values.containerSecurityContext "enabled" | toYaml | nindent 12 }}
+          {{- end }}
           volumeMounts: {{ include "docker-registry.volumeMounts" . | nindent 12 }}
       {{- if .Values.nodeSelector }}
       nodeSelector: {{ toYaml .Values.nodeSelector | nindent 8 }}
--- a/values.yaml
+++ b/values.yaml
@@ -152,6 +152,9 @@ configData:
       interval: 10s
       threshold: 3
 
+containerSecurityContext:
+  enabled: false
+
 securityContext:
   enabled: true
   runAsUser: 1000
--- a/templates/cronjob.yaml
+++ b/templates/cronjob.yaml
@@ -38,9 +38,7 @@ spec:
           priorityClassName: "{{ .Values.priorityClassName }}"
           {{- end }}
           {{- if .Values.securityContext.enabled }}
-          securityContext:
-            fsGroup: {{ .Values.securityContext.fsGroup }}
-            runAsUser: {{ .Values.securityContext.runAsUser }}
+          securityContext: {{ omit .Values.securityContext "enabled" | toYaml | nindent 12 }}
           {{- end }}
           containers:
             - name: {{ .Chart.Name }}
@@ -52,6 +50,9 @@ spec:
               - --delete-untagged={{ .Values.garbageCollect.deleteUntagged }}
               - /etc/docker/registry/config.yml
               env: {{ include "docker-registry.envs" . | nindent 16 }}
+              {{- if .Values.containerSecurityContext.enabled }}
+              securityContext: {{ omit .Values.containerSecurityContext "enabled" | toYaml | nindent 16 }}
+              {{- end }}
               volumeMounts: {{ include "docker-registry.volumeMounts" . | nindent 16 }}
           restartPolicy: OnFailure
           {{- if .Values.nodeSelector }}


schinmai-akamai:
+1 to this - would love to have the ability to pass securityContexts

joshsizer:
+1 this is a crucial feature

templates/deployment.yaml
```diff
@@ -42,11 +42,9 @@ spec:
...
       priorityClassName: "{{ .Values.priorityClassName }}"
       {{- end }}
-      {{- if .Values.securityContext.enabled }}
-      securityContext:
-        fsGroup: {{ .Values.securityContext.fsGroup }}
-        runAsUser: {{ .Values.securityContext.runAsUser }}
-      {{- end }}
+      {{- if .Values.securityContext }}
+      securityContext: {{ toYaml .Values.securityContext | nindent 8 }}
+        {{- end }}
```
joshsizer, line 47:
```suggestion
      {{- end }}
```

Mercbot7:
This is a needed feature for us as well!

erikfuego:
sorry everyone! i was on paternal leave. hoping this gets merged soon

joshsizer (approved):
LGTM

joshsizer:
@canterberry Could we get an `Approving review`?

This MR looks good to me, and it's backwards compatible. Also, happy to become a maintainer for this repo if you need some help going through these issues.

templates/deployment.yaml
```diff
@@ -82,6 +80,9 @@ spec:
               port: 5000
           resources: {{ toYaml .Values.resources | nindent 12 }}
           env: {{ include "docker-registry.envs" . | nindent 12 }}
+          {{- if .Values.containerSecurityContext }}
```
joshsizer, line 83:
```suggestion
          {{- if .Values.containerSecurityContext.enabled }}
```

values.yaml
```diff
@@ -153,7 +153,6 @@ configData:
       threshold: 3
 
 securityContext:
-  enabled: true
   runAsUser: 1000
   fsGroup: 1000
```
joshsizer, line 157:
```suggestion
  fsGroup: 1000
  
containerSecurityContext:
  enabled: false
```

joshsizer (changes requested):
@erikfuego I realize this is not backwards compatible as of now, if people have securityContext disabled, this update will now automatically enable it which may be undesired and breaking (since there are default fields for securityContext set).

In the case that someone already has it enabled, you receive the following warning when installing:

```
W0201 10:03:19.051240    7376 warnings.go:70] unknown field "spec.template.spec.securityContext.enabled"
```

I've suggested some ways to work around this by keeping the enabled field, and omitting it when populating the securityContext sections.

I was using the [Bitnami MongoDB chart](https://github.com/bitnami/charts/blob/4369898d264d231606e394b59ed4b3d926155d4c/bitnami/mongodb/templates/standalone/dep-sts.yaml#L79-L81) as a basis

erikfuego:
> @erikfuego I realize this is not backwards compatible as of now, if people have securityContext disabled, this update will now automatically enable it which may be undesired and breaking (since there are default fields for securityContext set).
> 
> In the case that someone already has it enabled, you receive the following warning when installing:
> 
> ```
> W0201 10:03:19.051240    7376 warnings.go:70] unknown field "spec.template.spec.securityContext.enabled"
> ```
> 
> I've suggested some ways to work around this by keeping the enabled field, and omitting it when populating the securityContext sections.
> 
> I was using the [Bitnami MongoDB chart](https://github.com/bitnami/charts/blob/4369898d264d231606e394b59ed4b3d926155d4c/bitnami/mongodb/templates/standalone/dep-sts.yaml#L79-L81) as a basis

Yeah thats a good point. Can you confirm that all the changes you suggested are in? The CI tests are failing. Im also going to test this locally with the values I set for the chart

erikfuego:
@joshsizer I reverted the changes you made to the `containerSecurityContext` because thats a new field I added to make the securityContext work for containers. And accepted your changes to leave securityContext along with the `enabled` field. The CI now passes

joshsizer:
> @joshsizer I reverted the changes you suggested for `containerSecurityContext` because thats a new field I added to make the securityContext work for containers. And accepted your changes to keep the `enabled` field for securityContext and omit it when populating securityContext

Awesome! I wonder, would it be good to keep the changes for containerSecurityContext, so that the schemas for each section are similar? I see either option as viable, but lean towards keeping them similar. I think lint was failing because we would have to add in to `values.yaml`

```yaml
containerSecurityContext:
  enabled: false
```

erikfuego:
> > @joshsizer I reverted the changes you suggested for `containerSecurityContext` because thats a new field I added to make the securityContext work for containers. And accepted your changes to keep the `enabled` field for securityContext and omit it when populating securityContext
> 
> Awesome! I wonder, would it be good to keep the changes for containerSecurityContext, so that the schemas for each section are similar? I see either option as viable, but lean towards keeping them similar. I think lint was failing because we would have to add in to `values.yaml`
> 
> ```yaml
> containerSecurityContext:
>   enabled: false
> ```

I prefer the pattern where `enabled` is not used because otherwise its an extra value to worry about. But if you want to use `enabled`, feel free to suggest the changes and I can apply them

erikfuego:
sorry about that weird merge, I merged this branch into a colleagues forked repo to test these changes, and for some reason his changes were merged into my branch.

erikfuego:
> The last change we need is to add to values.yaml:
> 
> ```yaml
> containerSecurityContext:
>   enabled: false
> ```

that broke the CI

joshsizer:
> > The last change we need is to add to values.yaml:
> > ```yaml
> > containerSecurityContext:
> >   enabled: false
> > ```
> 
> that broke the CI

Could you add into values.yaml

```
containerSecurityContext:
  enabled: false
```

That will fix the CI.

I'm not sure I can push to this branch

joshsizer (approved):
LGTM

@canterberry Can you review this as well?

erikfuego:
> LGTM
> 
> @canterberry Can you review this as well?

wanted to bumb this. Thanks!

joshsizer:
@erikfuego Would you be able to recommit with signed commits? I am now a maintainer, and can approve and merge, but this repository is setup to only allow signed commits

vyas-n (approved):

vyas-n:
@joshsizer

Not sure what the protocol should be moving forward, but I'm also a maintainer and I can now merge if you'd like.

joshsizer (approved):

joshsizer:
> @joshsizer
> 
> Not sure what the protocol should be moving forward, but I'm also a maintainer and I can now merge if you'd like.

Let's do it!

Issue #104, opened 2023-05-15
Garbage collection cronjob leaves registry in an inconsistent state


0xDEC0DE:
h1. Prerequisites

* `kubectl`
* `skopeo`
* A deployed registry configured with `persistence.deleteEnabled = true` and `garbageCollect.enabled = true`

# Steps to reproduce

## Setup
```
export REGISTRY=$NAME_OF_REGISTRY_INGRESS
export REGISTRY_POD=$NAME_OF_REGISTRY_POD_FROM_KUBECTL

docker login $REGISTRY
docker pull hello-world:latest
docker tag hello-world:latest ${REGISTRY}/hello-world:latest
docker push ${REGISTRY}/hello-world:latest
skopeo delete docker://${REGISTRY}/hello-world:latest
kubectl exec $REGISTRY_POD -- /bin/registry garbage-collect --delete-untagged=true /etc/docker/registry/config.yml
```

## Test
```
docker push ${REGISTRY}/hello-world:latest
docker pull ${REGISTRY}/hello-world:latest
```

# Expected result

Success

# Actual behavior

It fails, claiming layers already exist, etc.

# Workaround

Restarting the registry after garbage-collection makes it work as expected:
```
kubectl delete pod $REGISTRY_POD ; sleep 5  # XXX: restart registry
docker push ${REGISTRY}/hello-world:latest
docker pull ${REGISTRY}/hello-world:latest  # XXX: success
```

0xDEC0DE:
# Workaround

Disabling caching on the registry server will also make it behave correctly.  This can be achieved by setting `storage.cache.blobdescriptor` to an "invalid" value in the configs, e.g.:

```
docker-registry:
  configData:
    storage:
      cache:
        blobdescriptor: disabled
```

nabaruns:
Facing same issue.

BrianValente:
Can confirm, workaround works.

PR #105, opened 2023-05-20
Disable blob caching when garbage-collection is enabled

0xDEC0DE:
The garbage-collection cronjob will cause the registry to go into an inconsistent state if the blob cache is not cleared afterward.

Since the cronjob has no facility for doing that, disable blob caching entirely when garbage-collection is in use.

Fixes: Issue #104


nabaruns (approved):
LGTM.

PR #106, opened 2023-06-15
Adding Deployment Annotations

### LINKED ISSUE #86: Adding annotations for deployment to values
It would be very great to have annotations for the deployment itself. As there is currently no way to do that, we would have to clone and integrate the entire chart into our parent chart which is a mess just to change a few lines there.

For example the wave project for kubernetes requires a certain annotation on deployments / statefulsets in order to be active.
As there are annotations available for almost all other resources that would be a great minor addition which would help a lot!

ChevronTango:
Closes #86

This adds the ability for users to specify annotations on the deployment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ their default values.
 | `serviceAccount.create`     | Create ServiceAccount                                                                      | `false`         |
 | `serviceAccount.name`       | ServiceAccount name                                                                        | `nil`           |
 | `serviceAccount.annotations` | Annotations to add to the ServiceAccount                                                  | `{}`            |
+| `deployment.annotations`    | Annotations to add to the Deployment                                                       | `{}`            |
 | `service.port`              | TCP port on which the service is exposed                                                   | `5000`          |
 | `service.type`              | service type                                                                               | `ClusterIP`     |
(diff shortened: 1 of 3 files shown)


joshsizer (approved):
@ChevronTango This looks good to me.

@vyas-n What do you think?

joshsizer:
@ChevronTango Could you recommit these changes with signed commits? 

This comment has some resources to help with that if you need it: https://github.com/twuni/docker-registry.helm/pull/93#issuecomment-1979504730

vyas-n (approved):

vyas-n:
Looks good to me, just needs signed commits before it can be merged 👍🏽

ChevronTango:
Just pushed up a signed commit using Sigstore, however GitHub doesn't formally support those signatures yet, despite them being valid and secure. Hopefully that doesn't prevent the merge going ahead. If it does I'll resign using a GPG key.

joshsizer:
@ChevronTango Unfortunately GitHub is still saying that the commits need to be signed. Could you go ahead and commit them again with GPG/SSH?

ChevronTango:
@joshsizer that should be all ready to go now

joshsizer:
> @joshsizer that should be all ready to go now

Thank you!

PR #113, opened 2023-11-17
Support for setting the service name

perangel:
### Problem

It would be useful to be able to specify the service name and not rely on the `fullname` (which requires using `fullnameOverride` to set).

While `service.name` is defined in the `values.yaml`, it is not referenced anywhere. 

### Proposed Solution

- Set the default value of `service.name` to `""` and add an `if/else` conditional in the service template that will allow you to specify the service name.
- If unset, it will fallback to the existing behavior.

### DIFF
--- a/templates/service.yaml
+++ b/templates/service.yaml
@@ -1,7 +1,11 @@
 apiVersion: v1
 kind: Service
 metadata:
+{{- if .Values.service.name }}
+  name: {{ .Values.service.name }}
+{{- else }}
   name: {{ template "docker-registry.fullname" . }}
+{{- end }}
   namespace: {{ .Values.namespace | default .Release.Namespace }}
   labels:
     app: {{ template "docker-registry.name" . }}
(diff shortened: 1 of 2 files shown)


templates/service.yaml
```diff
@@ -1,7 +1,11 @@
 apiVersion: v1
 kind: Service
 metadata:
+{{- if .Values.service.name }}
+  name: {{ .Values.service.name }}
+{{- else }}
```
perangel, line 6: Alternatively, I think this could be written as:
```suggestion
  name: {{ .Values.service.name | default template "docker-registry.fullname" . }}
```

joshsizer (changes requested):
@perangel Nice catch, this is a much needed feature! I left a suggestion that I've tested locally and confirmed works for me.

If that suggestion looks good to you, please apply it and we can get this merged in 👍 

Edit: we do also need to make that change anywhere the service name is referenced, for example from the ingress resource.

PR #114, opened 2024-01-04
Fix backwards compatibility for k8s 1.24

ddelange:


.github/workflows/healm_publish.yaml
```diff
@@ -0,0 +1,29 @@
+name: helm-publish
```
ddelange, line 1: to be able to reference direct link to helm chart from release assets: https://github.com/ddelange/docker-registry.helm/releases/tag/v2.2.3

since we can release from any branch, this allowed me to test this PR on our live staging cluster

vyas-n (changes requested):
Hi @ddelange,

Your change for [templates/poddisruptionbudget.yaml](templates/poddisruptionbudget.yaml) makes sense, and I'd like to approve & merge it.

But you have an additional GitHub workflow [.github/workflows/healm_publish.yaml](.github/workflows/healm_publish.yaml) & Helm Chart version bump that I would like removed from this PR since it's not related to k8s 1.24 compatibility.

ddelange:
@vyas-n done

Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart for Docker Registry
 name: docker-registry
-version: 2.2.2
+version: 2.2.3
```
vyas-n, line 4: Still has the version bump, once this is removed it can be merged

vyas-n (changes requested):

Chart.yaml
ddelange: missed it, done

vyas-n (approved):
Thanks @ddelange!

PR #117, opened 2024-03-12
Update README.md to use correct default version

laverya:
The default tag has been 2.8.1 since the last release at the very least: https://github.com/twuni/docker-registry.helm/blob/v2.2.2/values.yaml#L22


joshsizer (approved):
@laverya Thanks for catching this!

PR #121, opened 2024-03-16
chore: add helm release action

joshsizer:
The chart-releaser-action requires that helm charts be placed into chart/<chart_name> to work. I've tried configuring it appropriately for this repo, but couldn't get it to work.

Instead, I'm using that action to install just the chart-releaser binary. Then, I'm using the `cr` binary to package, release, and update the index in the gh-pages branch.

Any commit on the main branch will cause a new release as long as the chart version has changed.

### DIFF
--- a/.github/workflows/helm_release.yaml
+++ b/.github/workflows/helm_release.yaml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
--- a/.github/workflows/helm_release.yaml
+++ b/.github/workflows/helm_release.yaml
@@ -1,9 +1,7 @@
 name: Release Charts
 
 on:
-  push:
-    branches:
-      - main
+  workflow_dispatch:
 
 jobs:
   release:
--- a/.github/workflows/helm_release.yaml
+++ b/.github/workflows/helm_release.yaml
@@ -1,0 +1,37 @@
+name: Release Charts
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  release:
+    permissions:
+      contents: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Configure Git
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+      
+      - name: Install chart-releaser
+        uses: helm/chart-releaser-action@v1.6.0
+        with:
+          install_only: true
+
+      - name: Run chart-releaser
+        env:
+          CR_TOKEN: "${{ secrets.CR_TOKEN }}"
+        run: |
+          owner=$(cut -d '/' -f 1 <<< "$GITHUB_REPOSITORY")
+          repo=$(cut -d '/' -f 2 <<< "$GITHUB_REPOSITORY")
+          cr package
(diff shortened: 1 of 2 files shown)


joshsizer:
@canterberry The release action needs a PAT that can write to the repo and create releases, stored in the actions secret called $CR_TOKEN. Would you be able to generate that and place it into the repository?

.github/workflows/helm_release.yaml
```diff
@@ -0,0 +1,37 @@
+name: Release Charts
+
+on:
+  push:
+    branches:
+      - main
+
```
vyas-n, line 7: Hey @joshsizer,

Mind changing the trigger from on push to `workflow_dispatch` instead?

I don't think I'd want every push to the repository's `main` branch to produce a release.

.github/workflows/helm_release.yaml
```diff
@@ -0,0 +1,37 @@
...
+jobs:
+  release:
+    permissions:
+      contents: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
```
vyas-n, line 15: Mind bumping this action up to `actions/checkout@v4`?

.github/workflows/helm_release.yaml
```diff
@@ -0,0 +1,37 @@
...
+        with:
+          fetch-depth: 0
```
vyas-n, line 17: `fetch-depth` can increase the time CI takes to clone the repo.

Does chart-releaser need the commit info to produce a release?

If not, I prefer if this input is removed.

vyas-n (changes requested):

.github/workflows/helm_release.yaml
joshsizer: @vyas-n A release is only produced when the version value in Chart.yaml changes on the main branch (not on every commit to main). Not sure if that changes your opinion at all, but if not I can definitely change it to `workflow_dispatch`.

.github/workflows/helm_release.yaml
joshsizer: @vyas-n I'm not completely sure on this. All the examples of using chart-releaser have had fetch-depth set to `0`. I can do some testing on my fork to confirm.

joshsizer: @vyas-n Tested with out `fetch-depth: 0` here and received an error:

https://github.com/joshsizer/docker-registry.helm/actions/runs/8429788969/job/23084576246#step:5:12

```shell
fatal: invalid reference: origin/gh-pages
Error: exit status [12](https://github.com/joshsizer/docker-registry.helm/actions/runs/8429788969/job/23084576246#step:5:13)8
```

vyas-n: Oh my bad, I didn't realize it used the checkout git repo directly to commit & push the changes to the `gh-pages` branch.

.github/workflows/helm_release.yaml
vyas-n: In this case, I'd still prefer that the workflow occurs on `workflow_dispatch`.

It'll keep the GitHub actions jobs a bit cleaner by only having executions when we are releasing, instead of each time it's pushed to the main branch and no-op'ing the commit.

vyas-n (approved):

canterberry:
> @canterberry The release action needs a PAT that can write to the repo and create releases, stored in the actions secret called $CR_TOKEN. Would you be able to generate that and place it into the repository?

I've generated a PAT and created the `CR_TOKEN` secret with that PAT as its value. This token will expire on March 28, 2025. As that date approaches, remind me and I'll replace it with a fresh one (or between now and then, we'll make sure at least one other person has read/write privileges for secrets on this repo).

PR #129, opened 2024-05-09
Wkbrd security patch

wkbrd:
This branch adds additional securityContext controls to align with those available from the Bitnami Helm charts.

This also renames the securityContext value to podSecurityContext in alignment with the Bitnami values.

Added Apache license details from Bitnami due to the README.md borrowing text from the Bitnami README.md file


canterberry (changes requested):
The core value proposition of this PR is the introduction of hardened default values for `securityContext` and `containerSecurityContext`.

The only part of this changeset that I believe to be strictly necessary to achieve this would be adjusting the default values in **values.yaml** (but without breaking interface compatibility by renaming `securityContext` to `podSecurityContext`).

This leaves the following changes to the `securityContext` and `containerSecurityContext` defaults in **values.yaml**:
```yaml
containerSecurityContext:
  enabled: true
  seLinuxOptions: {}
  allowPrivilegeEscalation: false
  capabilities:
    drop:
    - ALL
  privileged: false
  readOnlyRootFilesystem: true
  runAsUser: 1000
  runAsGroup: 1000
  runAsNonRoot: true
  seccompProfile:
    type: RuntimeDefault
securityContext:
  enabled: true
  fsGroupChangePolicy: Always
  sysctls: []
  supplementalGroups: []
  fsGroup: 1000
```

Are these defaults compatible with all currently officially supported versions of Kubernetes? If so, and given that these defaults should not interfere with overrides existing consumers of this chart may be using and depending on, I'd be comfortable approving this PR if it were distilled to the minimal changeset above (no changes to README or templates, no renaming of values which may be currently in use, no documentation of nested fields not directly used/defined/referenced by this chart).

canterberry:
Thanks @wkbrd! Always good to see some security-minded folks having a look and seeking to improve things.

I just took a review pass and left some feedback above. The gist: the new default values look good to me, but I'd like to see some diligence on compatibility so we don't inadvertently break things for folks when they upgrade this chart.

wkbrd:
Hi @canterberry , I think this addresses the changes you requested.  The only changed file is now values.yaml.

canterberry (approved):
:+1: LGTM! With the following caveats I leave to the other maintainers to discuss and consider:

The hardened default security contexts seem good to me, but since the default value of `containerSecurityContext.enabled` is changed from `false` to `true`, there's a possibility of unwanted/undesired side effects from having a container securityContext suddenly enabled after an upgrade. In particular, I could see this being a problem with permissions in deployments involving mounted volumes.

I'm also not an active *consumer* of this chart, so I can't speak to the compatibility of the new defaults with any given Kubernetes version or chart configuration, nor that any of the values in this changeset are necessarily accepted by Kubernetes or behave as they would seem.

I would *prefer* to have a secondary review on this, and/or a detailed report explaining the changes, so that the community is best equipped to understand what the defaults do and the rationale for each. Whether that is the burden of the author, the reviewer, or the project maintainers, is up for discussion.

PR #130, opened 2024-05-14
Update README.md

wkbrd:
This registry container does not have htpasswd in it. httpd does. Recommend changing so that documentation is usable.


canterberry (approved):

canterberry:
For reference, here's what `httpd:2` currently resolves to: https://hub.docker.com/layers/library/httpd/2/images/sha256-9ff6162ebb46e14a4846b88aed4c3ac67b8a61a80fdafe29b2f1714c70c5a101

Issue #135, opened 2024-06-11
config path will change with distribution 3


mawatech:
With the next release the config path will change from `/etc/docker/registry/config.yml` to `/etc/distribution/config.yml` 

See https://github.com/distribution/distribution/pull/4365

Perhaps it make sense to make the mount path configurable?

canterberry:
Thanks for bringing this up, @mawatech!

That will definitely be a breaking change, since that mount path is [hard-coded](https://github.com/twuni/docker-registry.helm/blob/main/templates/_helpers.tpl#L151
) right now.

Are there other breaking changes to consider in v3?

This seems like a two-part fix:

 1. Update the current version of the chart to support an override to the configuration path (i.e: introduce a new configuration value and wire that into the volumeMounts helper), using a default equal to the currently hard-coded value so there are no functional changes for existing users of this chart.
 2. When v3 is officially released, update the default `image.tag` and config path value (+ any other necessary adjustments). Whether that's a major version increment for this chart as well, I defer for discussion.

joshsizer:
> 2. When v3 is officially released, update the default `image.tag` and config path value (+ any other necessary adjustments). Whether that's a major version increment for this chart as well, I defer for discussion.

I would argue for a major chart version increment for two reasons - the underlying image is a major change, and as well we are introducing breaking changes to the chart

Mercbot7:
UPDATE: `v3.0.0` has been officially released: https://github.com/distribution/distribution/releases/tag/v3.0.0

Mercbot7:
@mawatech, @joshsizer @canterberry  Am I wrong in that there are three total places this is hardcoded?

https://github.com/twuni/docker-registry.helm/blob/791c161883f7b1b8f471d045407b3a2575db3365/templates/cronjob.yaml#L61

https://github.com/twuni/docker-registry.helm/blob/791c161883f7b1b8f471d045407b3a2575db3365/templates/deployment.yaml#L63

https://github.com/twuni/docker-registry.helm/blob/791c161883f7b1b8f471d045407b3a2575db3365/templates/_helpers.tpl#L151

PR #142, opened 2024-08-23
make the registry mount point configurable

mawatech:
Fix for #135 

This makes the config mount point configurable.


Clovel:
Any news on this ?

PR #143, opened 2024-08-28
use mirrored registry from republic.ecr.aws to avoid Docker Hub rate limit

kahirokunn:


taraszka (commented):
I think this won't work, unless you are at AWS, no?

kahirokunn:
This is available from anywhere without restrictions

taraszka:
> This is available from anywhere without restrictions

Are you sure? I am getting:
```
❯ docker pull public.ecr.aws/docker/library/registry
Using default tag: latest

What's next:
    View a summary of image vulnerabilities and recommendations → docker scout quickview public.ecr.aws/docker/library/registry
Error response from daemon: pull access denied for public.ecr.aws/docker/library/registry, repository does not exist or may require 'docker login': denied: Your authorization token has expired. Reauthenticate and try again.
```

it should not ask me for any credentials if it's public.

joshsizer:
@kahirokunn Thanks for taking the time to contribute!

I am going to close this as this is already an overridable value - if you would like to use `public.ecr.aws/docker/library/registry` you are free to do so in your own deployment.

PR #159, opened 2025-03-15
Add option to conditionally create Service resource     

bwdmr:
- Add new parameter `service.create` (default: false) to control Service creation
    - Wrap Service template in conditional to respect the new parameter
    - Update documentation and default values to reflect this change
    
    This allows users who don't need the Service resource (e.g., when using
    Ingress directly or external service management) to skip its creation.


joshsizer:
Hi @bwdmr, could you please split this PR into two? I see two changes here - the optional creation of the service & creating the `docker-registry.namespace` template.

Also, please make sure to sign your commits! Otherwise we cannot merge changes in.

PR #164, opened 2025-04-14
Update garbageCollect cronjob

Mercbot7:
This change will provide `podLabels` and `podAnnotations` options specific to the `garbageCollect` cronjob job pod with defaults to the root `podLabels` and `podAnnotations`. There are a few minor fixes for yaml rendering for the cronjob as well where spaces were left in the cronjob yaml. There are `podLabels` and `podAnnotations` options added in the `values.yaml` file and some minor spacing fixes in that file as well.

I used the [merge function](https://helm.sh/docs/chart_template_guide/function_list/#merge-mustmerge) in Helm along with [default](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function) and an empty [`dict`](https://helm.sh/docs/chart_template_guide/function_list/#dict) (dictionary). The logic and functions should prefer the `garbageCollect.podLabels/podAnnotations` values over the root `podLabels/podAnnotations` and if neither are provided should output an empty map.

### DIFF
--- a/templates/cronjob.yaml
+++ b/templates/cronjob.yaml
@@ -31,11 +31,11 @@ spec:
           labels:
             release: {{ .Release.Name }}
             {{- if or .Values.podLabels .Values.garbageCollect.podLabels }}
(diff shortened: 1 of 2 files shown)


joshsizer (approved):
LGTM

PR #168, opened 2025-06-03
[Fixes #135] Updated chart to accept configPath to fix distribution's 3.0.0 breaking change

### LINKED ISSUE #135: config path will change with distribution 3

Clovel:
Fixes #135

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,7 @@ their default values.
 | `secrets.swift.password`    | Password for Swift configuration                                                           | `nil`           |
 | `secrets.haSharedSecret`    | Shared secret for Registry                                                                 | `nil`           |
 | `configData`                | Configuration hash for docker                                                              | `nil`           |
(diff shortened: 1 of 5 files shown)


mawatech:
i would say duplicate of #142

Clovel:
> i would say duplicate of #142 

It is, but the commits are signed-off and GPG signed. I understand that that's a requirement for PRs to be accepted here so I took the liberty of doing it again with signed commits. 

Also ran a lot of tests with my version of the chart in staging and production.

joshsizer (approved):
Tested this on my own deployment of docker registry, I am still able to pull and push images

PR #172, opened 2025-06-09
Upgrade to Registry v3.0.0

zatkowich:


zatkowich:
I see this already has been implemented :-)

PR #173, opened 2025-06-14
Use labels suggested by K8s and Helm best practices

pinkavaj:
This improves the compatibility and makes it easier to work with various tools, which take advantage from the assumption of existence of those labels.

    https://helm.sh/docs/chart_best_practices/labels/
    https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

Changing the selectors is breaking change, so the bump of the major version of the helm chart.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -23,6 +23,20 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- end -}}
 {{- end -}}
 
+{{- define "docker-registry.match-labels" -}}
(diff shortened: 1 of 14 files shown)


Mercbot7:
I like what this does but do have to ask. Why does it need to be a breaking change? can the labels not co-exist for a time and a deprecation notice be given?

pinkavaj:
> I like what this does but do have to ask. Why does it need to be a breaking change? can the labels not co-exist for a time and a deprecation notice be given?

Sure they can, NP here.

PR #176, opened 2025-07-22
Image tag set to '' now leads to `image: "registry"` instead of image: `"registry:"`

ClementJ35:
Hi,

We need to use a non-tagged image as it's shipped in our nodes filesystem so updates of nodes doesn't require a synchronized update of deployment as we have to follow the shipped-only image.
With this PR, settings explicitly an empty value for image.tag also remove the leading `:`.
With any other values, functionality remains the same.

### DIFF
--- a/templates/deployment.yaml
+++ b/templates/deployment.yaml
@@ -55,7 +55,7 @@ spec:
       {{- end }}
       containers:
         - name: {{ .Chart.Name }}
-          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          image: "{{ .Values.image.repository }}{{ if not (empty .Values.image.tag) }}:{{ .Values.image.tag }}{{ end }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           command:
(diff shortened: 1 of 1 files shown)


joshsizer:
Hi @ClementJ35, have you tried setting .Values.image.tag to `latest` in your use case?

ClementJ35:
Hi @joshsizer ,
No I didn't... And it works... :sleepy: 
In my head latest was an explicit tag that must exist on the repo. But using it with the `imagepullpolicy: Never` is working as intended.
As nobody else seems to react to my issue and PR, I'll close them.
Thanks for that and sorry wasting your time.

joshsizer:
@ClementJ35 Glad to hear you got it working on your side. In my experience "untagged" images are really just shorthand for `latest` tag (ie, docker clients assume you mean latest if you don't include a tag). Images that have had all their tags (including latest) deleted become dangling, so they aren't really accessible anymore. BTW, not a waste of time at all. Always looking to help support the users of this chart.

